PR #24, opened 2023-02-01
Files Added in

Anooj-Pai:
I got the files to read in from our test folder and added some extra files to make sure that it gets them all.


cli.js
```diff
@@ -6,9 +6,14 @@ import json1 from 'ot-json1';
 import { WebSocketServer } from 'ws';
 import WebSocketJSONStream from '@teamwork/websocket-json-stream';
 import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
```
curran, line 10: Niiiice!

src/App.jsx
```diff
@@ -17,10 +18,96 @@ function App() {
     });
   }, []);
 
+
+  function showFiles() {
+    let arrow = document.querySelectorAll(".arrow");
```
curran, line 23: Prefer const

src/App.jsx
```diff
@@ -17,10 +18,96 @@ function App() {
...
+    for (var i = 0; i < arrow.length; i++) {
```
curran, line 24: We'll change this up to do it "the React way", but this is great as a first pass.

src/App.jsx
```diff
@@ -17,10 +18,96 @@ function App() {
...
+      parsedData[i] = {
+        name: file,
+        content: content,
+      }
+      i++;
+    }
+  }
+
+  function showContent() {
+    let file = document.querySelectorAll(".sub-menu li");
```
curran, line 57: Yeah this is definitely not the React way. We'll clean this up too.

src/App.jsx
```diff
@@ -17,10 +18,96 @@ function App() {
...
+  fillData();
+  showContent();
+
+
   return (
-    <div className="App">
-      <pre>{JSON.stringify(data, null, 2)}</pre>
-    </div>
+    <>
+      <link href='https://unpkg.com/boxicons@2.0.7/css/boxicons.min.css' rel='stylesheet'></link>
```
curran, line 80: This CSS import should not go here. Please move it into `index.html`.

cli.js
```diff
@@ -6,9 +6,14 @@ import json1 from 'ot-json1';
...
 
 const fullPath = process.cwd();
 
+console.log(fullPath);
```
curran, line 14:
```suggestion
```

src/App.jsx
```diff
@@ -17,10 +18,96 @@ function App() {
...
+
+  function showFiles() {
+    let arrow = document.querySelectorAll(".arrow");
+    for (var i = 0; i < arrow.length; i++) {
+      arrow[i].addEventListener("click", (e) => {
+        let arrowParent = e.target.parentElement.parentElement;//selecting main parent of arrow
+        arrowParent.classList.toggle("showMenu");
+        let folder = document.getElementById("folderIcon");
+        if (folder.className === "bx bx-folder-open") {
+          folder.className = "bx bx-folder";
```
curran, line 30: This rendering logic should be moved into the JSX.

src/App.jsx
```diff
@@ -17,10 +18,96 @@ function App() {
...
+          folder.className = "bx bx-folder-open";
+        }
+      });
+    }
+  }
+
+  let parsedData = {};
+
+  function fillData() {
+    let fileList = document.querySelector(".sub-menu");
```
curran, line 41: It's generally a bad idea to query the DOM in the React world. It's better to generate the DOM based on our data.

curran (changes requested):
A few nits. Maybe some of them we can work through together, but I think some of them you can handle. Great work!

PR #36, opened 2023-02-07
Added All

Anooj-Pai:
Added lang support
Added Tab

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -10,6 +10,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 
 const fullPath = process.cwd();
+console.log(fullPath);
 
 // Isolate files, not directories.
 // Inspired by https://stackoverflow.com/questions/41472161/fs-readdir-ignore-directories
@@ -19,7 +20,6 @@ const files = fs
   .map((dirent) => dirent.name);
 
 const initialDocument = {};
-console.log(files);
 files.forEach((file) => {
   const id = Math.floor(Math.random() * 10000000000);
   initialDocument[id] = {
@@ -28,14 +28,13 @@ files.forEach((file) => {
   };
 });
 
-console.log(initialDocument);
+//console.log(initialDocument);
 
-console.log('Welcome to VZCode!');
 
 ShareDB.types.register(json1.type);
 
 const app = express();
-const port = 3030;
+const port = 5173;
 
 const shareDBBackend = new ShareDB();
 const shareDBConnection = shareDBBackend.connect();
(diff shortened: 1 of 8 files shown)


src/App.jsx
```diff
@@ -63,6 +64,7 @@ function App() {
                 ? Object.keys(data).map((key) => (
                   <li onClick={() => {
                     setActiveFileId(key)
+                    setTabName(data[key].name)
```
curran, line 67: I don't like this, because it duplicates state.

Think about it like this: `activeFileId` and `tabName` _always_ change together.

I would suggest to just use `activeFileId` to inform `tabName` by computing a derived value rather than introducing a new piece of state.
```suggestion
```

src/App.jsx
```diff
@@ -15,6 +15,7 @@ function App() {
 
   const [isFileMenuOpen, setIsFileMenuOpen] = useState(false);
   const [activeFileId, setActiveFileId] = useState(null)
+  const [tabName, setTabName] = useState(null)
```
curran, line 18: No need for the `useState` call, just derive it from what's already there.
```suggestion
  const tabName = (data && activeFileId) ? data[activeFileId].name : "";
```

src/style.css
```diff
@@ -10,10 +10,24 @@
...
   font-family: 'Poppins', sans-serif;
 }
 
+.tab {
+  width: max-content;
+  height: 30px;
+  background: rgb(48, 48, 48);
+  text-align: center;
+  color: #fff;
+  font-family: ui-monospace, SFMono-Regular, SF Mono, Menlo, Consolas, Liberation Mono, monospace;
```
curran, line 23: Opportunity to refactor CSS so that this list only appears once.

curran (changes requested):
Nice work!

I suggest to remove that new `useState` call, as it's pretty much duplicating the active file state. Specific code changes in comments.

cli.js
```diff
@@ -10,6 +10,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 
 const fullPath = process.cwd();
+console.log(fullPath);
```
curran, line 13:
```suggestion
```

src/App.jsx
```diff
@@ -6,6 +6,7 @@ import "./style.css";
 
 const { Connection } = ShareDBClient;
 const socket = new WebSocket("ws://" + window.location.host);
+console.log(window.location.host);
```
curran, line 9:
```suggestion
```

src/App.jsx
```diff
@@ -27,11 +28,17 @@ function App() {
     });
   }, []);
 
-  const activeFileText = (data && activeFileId) ? data[activeFileId].text : "";
-
+  function close() {
+    console.log('close')
```
curran, line 32:
```suggestion
```

curran (approved):
Awesome work!

PR #43, opened 2023-02-13
No backspace auto save

Anooj-Pai:
This version works when you dont remove anything from the file, on enter it will save the line but if you try and remove a whole line or more it adds some objects instead

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -18,7 +18,6 @@ const files = fs
   .filter((dirent) => dirent.isFile())
   .map((dirent) => dirent.name);
 
-// Compute our initial document by reading the file system.
 const initialDocument = {};
 files.forEach((file) => {
   const id = Math.floor(Math.random() * 10000000000);
@@ -28,12 +27,11 @@ files.forEach((file) => {
   };
 });
 
-console.log('Welcome to VZCode!');
 
 ShareDB.types.register(json1.type);
 
 const app = express();
-const port = 5173;
+const port = 3030;
 
 const shareDBBackend = new ShareDB();
 const shareDBConnection = shareDBBackend.connect();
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -53,6 +53,7 @@
     "react-dom": "^18.2.0",
     "sharedb": "^3.2.1",
     "sharedb-client-browser": "^2.2.49",
+    "vite-plugin-fs": "^1.0.0-beta.5",
```
curran, line 56: Why do we need this plugin?
```suggestion
```

src/App.jsx
```diff
@@ -1,6 +1,7 @@
 import { useState, useEffect, useMemo } from "react";
 import ShareDBClient from "sharedb-client-browser/sharedb-client-json1-browser.js";
 import { CodeEditor } from './CodeEditor';
+import fs from 'fs';
```
curran, line 4: The front end does not need to know about the file system, as that will only be interfaced with on the server.
```suggestion
```

cli.js
```diff
@@ -44,17 +42,37 @@ wss.on('connection', (ws) => {
...
 const __dirname = path.dirname(__filename);
 const dir = path.join(__dirname, '/dist');
 
 app.use(express.static(dir));
 
-// Create initial document
 const shareDBDoc = shareDBConnection.get('documents', '1');
 shareDBDoc.create(initialDocument, json1.type.uri);
 
+shareDBDoc.subscribe(() => {
```
curran, line 59: Nice as a first pass!

I think it can be simplified such that we don't need to look at the ops at all.

Also, eventually we should debounce. Currently this will save the file on each keystroke, which is too frequent.

Here's a sketch of an idea:

 * On each op, set a `dirty` flag and start a timer
 * On each op while the timer is going, reset the timer
 * When the timer finishes, write only the files that changed
 * We can detect which files have changed by storing in memory the state of the files as they were the last time a save happened and comparing each one

curran (approved):
Great as a first pass!

PR #47, opened 2023-02-20
Rename Files

curran:
Changes:

 * Fix proxy setup with Vite dev server (it was actually serving the `dist` build earlier but I didn't notice)
 * Disable text selection on double click for file name in sidebar
 * Fix React warnings re:keys
 * Implement rename of files in the client
 * Implement rename of files in the server


curran:
Merging to cut a release.

PR #51, opened 2023-03-04
Make JSON1 with presence available to frontend and backend

curran:
Summary of changes:
 * Copy source into `src` from https://github.com/ottypes/json1/pull/47
 * De-conflict the naming so that we can use this OT type with our existing ShareDB client build


curran:
Reached a juncture here:
 * Problem: our pure Node.js `cli.js` script cannot directly import the source tree from JSON1 with presence because it's TypeScript.
 * Solution space: we need to somehow transpile that TypeScript into JavaScript for it to run in Node
 * Option A: Bring in the full JSON1 source tree including its `package.json`, which contains a build script that outputs JS that we _can_ import directly in our Node script.
 * Option B: Introduce a build step for our CLI script that uses Vite, which does support TypeScript
 * Option C: Adopt Vite in our CLI to dynamically import the TypeScript JSON1 implementation.

Considering the endgame scenarios as context:
 * Either the presence PR is eventually merge into JSON1, or we create and maintain a fork of JSON1 that gets published to NPM
 * In either of these scenarios, our code will most resemble going with option A above.

Therefore I'm leaning towards option A.

curran:
Tried option A to no avail.

The server side works, but the client side has issues with the build. Even after adding the CommonJS Vite plugin (which I really would rather not do), the build fails as `process` is undefined (the Node.js way of getting at environment variables).

It occurred to me that the best way forward might be to isolate all the legacy build madness and add JSON1 with presence to https://github.com/vizhub-core/sharedb-client-browser . That way, nothing needs to change about our code here except to import another asset from that package.

PR #53, opened 2023-03-08
Presence First Pass

### LINKED ISSUE #16: Presence cursors
Cursors that show the other users' positions and selections.

Maybe each participant gets their own color?

Here's how VSCode + LiveShare does it:

Image

The cursors should probably disappear if a user is idle for a long time.

curran:
Closes #16

image

First pass of the feature only.

Summary of changes:
 * Set up ShareDB `LocalPresence` instance
 * Listen for changes in CodeMirror
 * Broadcast presence using ShareDB API
 * Display the remote presence
 * Tested multiple connections (had to pass the `sort` arg to `Decoration.set` to fix this)
 * Tested editing after receiving presence. Had to add a `setTimeout` before dispatching the change to fix this, so there are not synchronous nested CodeMirror updates. This feels like somewhat of a hack, but not sure a better way.
 * Tested that presence cursors are transformed correctly when editing
 * Tested that presence cursors work across newlines

Current limitations:
 * Everyone's cursor is yellow
 * Does not handle removing presence yet
 * Does not yet support range selections (only the start of the selection is displayed remotely, for now)

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -18,6 +18,11 @@ const files = fs
   .filter((dirent) => dirent.isFile())
   .map((dirent) => dirent.name);
 
+// Initialize the document using our data structure for representing files.
+//  * Keys are file ids, which are random numbers.
+//  * Values are objects with properties:
+//    * text - the text content of the file
+//    * name - the file name
 const initialDocument = {};
 files.forEach((file) => {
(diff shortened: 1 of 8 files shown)


Anooj-Pai (approved):
This looks great! Everything installs and runs perfectly. May change some styling but great for now!

PR #54, opened 2023-03-10
New sidebar

Anooj-Pai:
Created the new sidebar based off of the vizhub v2 one. Everything works except when you hover over for the edit and delete they show up for each file instead of just the one that your are hovering over. This probably has a simple fix that I am overlooking but I will try to fix it. The rename works but the delete button still needs functionality.


cli.js
```diff
@@ -76,9 +76,11 @@ const save = () => {
   // Take a look at each file that we have currently.
   for (const key of Object.keys(currentDocument)) {
     // Handle changing of text content.
-    if (previousDocument[key].text !== currentDocument[key].text) {
-      const { name, text } = currentDocument[key];
-      fs.writeFileSync(name, text);
+    if (previousDocument[key] && currentDocument[key]) {
```
curran, line 79: Nice guard.

cli.js
```diff
@@ -90,6 +92,31 @@ const save = () => {
 
     // TODO handle creating files
     // TODO handle deleting files
+    const oldKeys = Object.keys(previousDocument);
+    const newKeys = Object.keys(currentDocument);
+    const removedKeys = oldKeys.filter((key) => !newKeys.includes(key));
+    removedKeys.forEach((key) => {
+      fs.unlinkSync(fullPath + '/' + previousDocument[key].name, (err) => {
+        process.on(err, () => { });
```
curran, line 100: This does not look like correct usage of `process.on`. See https://nodejs.org/api/process.html#process_process_events

I would recommend that we just throw the error to surface it.
```suggestion
        if (err) throw err;
```

cli.js
```diff
@@ -90,6 +92,31 @@ const save = () => {
 
     // TODO handle creating files
     // TODO handle deleting files
+    const oldKeys = Object.keys(previousDocument);
+    const newKeys = Object.keys(currentDocument);
+    const removedKeys = oldKeys.filter((key) => !newKeys.includes(key));
+    removedKeys.forEach((key) => {
+      fs.unlinkSync(fullPath + '/' + previousDocument[key].name, (err) => {
```
curran, line 99: Interesting that this uses `fullPath`, but the `writeFileSync` call worked fine without this.

Maybe it's not required?

```suggestion
      fs.unlinkSync(previousDocument[key].name, (err) => {
```

cli.js
```diff
@@ -90,6 +92,31 @@ const save = () => {
...
+        process.on(err, () => { });
+      });
+    });
+
+    const addedKeys = newKeys.filter((key) => !oldKeys.includes(key));
+    console.log(addedKeys);
```
curran, line 105:
```suggestion
```

cli.js
```diff
@@ -90,6 +92,31 @@ const save = () => {
...
+    addedKeys.forEach((key) => {
+      let { name, text } = currentDocument[key];
```
curran, line 107:
```suggestion
      const { name, text } = currentDocument[key];
```

cli.js
```diff
@@ -90,6 +92,31 @@ const save = () => {
...
+      if (!text) {
+        text = "";
+      }
+      fs.writeFileSync(name, text, (err) => {
+        if (err) {
+          if (err.code === 'ENOENT') {
```
curran, line 113: Does this happen? I would suggest we handle errors consistently, maybe just by throwing them for now.

src/App.jsx
```diff
@@ -119,11 +121,41 @@ function App() {
     [shareDBDoc]
   );
 
+  const deleteFile = useCallback(
+    (key) => {
+      const currentDocument = shareDBDoc.data;
+      const nextDocument = { ...currentDocument };
+      delete nextDocument[key];
+      shareDBDoc.submitOp(diff(currentDocument, nextDocument));
```
curran, line 129: Nice.

src/App.jsx
```diff
@@ -119,11 +121,41 @@ function App() {
...
+    if (!newName) return;
+    const currentDocument = shareDBDoc.data;
+    const nextDocument = {
+      ...currentDocument,
+      [randomId()]: {
+        name: newName,
+        content: '',
+      },
+    };
+    shareDBDoc.submitOp(diff(currentDocument, nextDocument));
```
curran, line 146: Nice. We may want to consider opening the new file also, immediately after creating it. I think VSCode does this.

src/App.jsx
```diff
@@ -119,11 +121,41 @@ function App() {
...
+  }, [shareDBDoc]);
+
   // True if we are ready to actually render the active tab.
   const tabValid = data && activeFileId;
+  const [utils, setUtils] = useState(false);
+  const [settings, setSettings] = useState(false);
 
   return (
     <>
+      <div className='settingsPosition'>
```
curran, line 156: Consistent naming
```suggestion
      <div className='settings-position'>
```

src/App.jsx
```diff
@@ -146,67 +178,58 @@ function App() {
...
-                </a>
+              <li className='files'>
+                <div className='full-Box'>
+                  <div>
+                    <a className='link-name' href="#">
+                      Files
+                    </a>
+                  </div>
+                  <div>
+                    <i class='bx bxs-file-plus newBTN' style={{ color: '#dbdde1' }} onClick={() => { createFile() }} ></i>
```
curran, line 191: Simplify
```suggestion
                    <i class='bx bxs-file-plus newBTN' style={{ color: '#dbdde1' }} onClick={createFile} ></i>
```

src/App.jsx
```diff
@@ -146,67 +178,58 @@ function App() {
...
                 ? Object.keys(data).map((key) => (
-                    <li
-                      key={key}
-                      onClick={() => {
-                        setActiveFileId(key);
-                        if (!tabList.includes(key)) {
-                          setTabList([...tabList, key]);
-                        }
+                  < li className="file" >
+                    <div className='full-Box' onClick={() => {
```
curran, line 198: Would be good to pull this out into a `useCallback`.

src/App.jsx
```diff
@@ -146,67 +178,58 @@ function App() {
...
                 : null}
             </ul>
           </li>
           <li>
-            <div className="profile-details">
+            <div className="settings">
               <a href="#">
-                <i className="bx bx-cog"></i>
-                <span className="link-name">Setting</span>
+                <span className='settings' onClick={() => setSettings(!settings)}>Settings</span>
```
curran, line 227: The parent `div` and child `span` have the same class, which might lead to some confusion.

curran (approved):
Looks amazing! I have a few nits for the code, but overall really great progress!

src/style.css
```diff
@@ -98,113 +98,85 @@ body {
...
+  align-items: center;
+  justify-content: center;
+  flex-direction: column;
 }
 
-.sidebar .profile-details {
-  position: fixed;
-  bottom: 0;
-  width: 200px;
+.settingsPosition {
```
curran, line 178:
```suggestion
.settings-position {
```

PR #57, opened 2023-03-13
Sidebar Cleanup

curran:
Changes:
 * Run Prettier
 * Refactor
 * Fix various bugs and warnings
 * Remove incorrect Node API usage (the `sync` variants do not accept a callback at all, they just throw if there's an error)
 * Make delete file work woohoo!
 * Show utils per file on hover


src/App.jsx
```diff
@@ -123,25 +150,38 @@ function App() {
 
   const deleteFile = useCallback(
     (key) => {
+      console.log('deleting file');
```
curran, line 153:
```suggestion
```

Anooj-Pai (approved):
This is so much better! The back end is much simpler now!

PR #58, opened 2023-03-13
Adopt Bootstrap

curran:
I'd like to adopt [Bootstrap](https://getbootstrap.com/docs/5.3/getting-started/introduction/) for our UI elements in the Settings panel. It's a fantastic library, and we can theme it later as we see fit. They have solved tons of really hard UI problems like how to do modals and many more.

I got it working! Here's what the new Settings modal looks like:

image

(content totally subject to change, just a placeholder for now)

However, the Bootstrap CSS unfortunately interferes with the styling of our existing elements:

image

I started on fixing these, but realized it would likely create merge conflicts with https://github.com/vizhub-core/vzcode/pull/57, so I think I'll hold off on these fixes until that gets merged. Most of the styling interference can probably be solved by:

 * Changing `<ul>` and `<li>` to `<div>`
 * Removing usage of `<a>` in the sidebar


curran:
Current status: all good! Ready for review @Anooj-Pai 

image

image

Anooj-Pai:
Looks good, only thing left is to implement the change of the theme in the back end and refresh the instance.

Issue #63, opened 2023-03-17
Release as usable product


curran:
I'd like to cut a release that folks can try out.

For this, we should disable partially working features, specifically the settings panel.

Idea: put it behind a "feature flag" that is enabled in development but disabled when running the production/release build.

PR #67, opened 2023-03-21
Feature flag for release

curran:
Disables the settings box for release mode only. The idea is to release a complete product, and we can bring the settings panel into the release once it is fully operational.

Towards #63


Anooj-Pai (commented):
Good idea and good implimentation

PR #75, opened 2023-03-23
Change debounce

Anooj-Pai:


src/settings.jsx
```diff
@@ -74,11 +74,30 @@ const themes = [
   { value: 'zenburn', label: 'zenburn' },
 ];
 
+const saveTimes = [
+  { value: 1, label: '1 second' },
+  { value: 5, label: '5 seconds' },
+  { value: 10, label: '10 seconds' },
+  { value: 30, label: '30 seconds' }
```
curran, line 81: TBH I don't think anyone would want to wait 30 seconds to save their code.

The range might make more sense between, like 0.5 seconds and 2 seconds.

Also, we may want to consider adding an option to not auto-save at all, and allow users to manually trigger saving like traditional IDEs.

curran (commented):
Nice!

PR #85, opened 2023-03-25
Support Ngrok

Anooj-Pai:


cli.js
```diff
@@ -10,6 +10,8 @@ import { fileURLToPath } from 'url';
 import { json1Presence } from './src/ot.js';
 import { randomId } from './src/randomId.js';
 import open from 'open';
+import ngrok from 'ngrok';
```
curran, line 13: I believe it should be possible for users to leverage NGrok without VZCode having any direct dependency on it.

cli.js
```diff
@@ -129,5 +132,9 @@ shareDBDoc.subscribe(() => {
 
 server.listen(port, () => {
   console.log(`Editor is live at http://localhost:${port}`);
-  open(`http://localhost:${port}`);
+  (async function () {
+    await ngrok.authtoken('2NSu2nKIuMoHIc5z9XDPIgm2gpz_7oDwUferEeadw1Mvc66Qi');
```
curran, line 136: This API key is, I would guess, tied to your personal NGrok account? I would suggest we do not add this to the repo. If there is a strong need to use an API key, it should probably go into an environment variable.

localhost.crt
curran, line 1: Generally the `.key` and `.crt` files should not go in a repository, as these are environment-specific (tied to the particular NGrok account).

package.json
```diff
@@ -33,7 +33,10 @@
     "Data",
     "Visualization"
   ],
-  "author": "Curran Kelleher",
+  "authors": [
+    "Curran Kelleher",
+    "Anooj Pai"
```
curran, line 38: Nice change! It would be best to make a separate PR so this can get in, independently of the NGrok stuff.

package.json
```diff
@@ -51,7 +54,9 @@
     "codemirror-ot": "^4.0.0",
     "diff-match-patch": "^1.0.5",
     "express": "^4.18.2",
+    "https": "^1.0.0",
```
curran, line 57: I don't think we need this dependency.

src/App.jsx
```diff
@@ -252,42 +253,42 @@ function App() {
...
-                              renameFile(key);
-                            }}
-                          ></i>
-                          <i
-                            className="bx bx-trash"
-                            style={{ color: '#eb336c' }}
-                            onClick={handleDeleteFileClick(key)}
-                          ></i>
-                        </div>
+                  <div className="file" key={key}>
```
curran, line 256: This seems to be a Prettier-related change. I wonder what is going on with this. Maybe if we in introduce a `.prettierrc`, this issue might go away. That would the make VSCode Prettier plugin (which might be the source of this particular change) do the same thing as our `npm run prettier` script.

curran (changes requested):
Let's put this one on hold for a bit.

Generally the `.key` and `.crt` files should not go in a repository.

I believe it should be possible for users to leverage NGrok without VZCode having any direct dependency on it.

Researching a bit,

>You can create basic http-https-tcp tunnel without an [authtoken](https://ngrok.com/docs#authtoken). For custom subdomains and more you should obtain an authtoken by [signing up at ngrok.com](https://ngrok.com/).

I would rather leave it up to users if they want to set up a custom subdomain. I think for most use cases, a basic http-https-tcp tunnel would be sufficient.

Anooj-Pai:
@curran can you call to talk about some of these changes?

PR #91, opened 2023-03-28
adds Ngrok compatibility

Anooj-Pai:
Based on an environmental  variable that the user sets the session will be served on either localhost or an Ngrok server tunnel


curran:
Very nice!

PR #93, opened 2023-03-28
add documentation

Anooj-Pai:
added directions of how to run with it installed globally but not sure how were working with the env vars yet so left that as coming soon


curran (approved):
Awesome!

PR #94, opened 2023-04-04
File listening

Anooj-Pai:
I have it listening to the directory but when I try send the op it gives a mismatch error. I know you know how to fix this so its just figuring out how to send the data to sharedb now.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -12,6 +12,9 @@ import { randomId } from './src/randomId.js';
 import open from 'open';
 import ngrok from 'ngrok';
 import dotenv from 'dotenv';
+import { diff } from './src/diff.js';
+import chokidar from 'chokidar';
+import { useCallback } from 'react';
 
 // The time in milliseconds by which auto-saving is debounced.
 const autoSaveDebounceTimeMS = 800;
@@ -49,11 +52,6 @@ files.forEach((file) => {
 ShareDB.types.register(json1Presence.type);
 
 const app = express();
-
-app.post('/saveTime', (req, res) => {
-  //autoSaveDebounceTimeMS = req.body.autoSaveDebounceTimeMS;
-  console.log('autoSaveDebounceTimeMS', req.body);
-});
 
 // Use ShareDB over WebSocket
 const shareDBBackend = new ShareDB({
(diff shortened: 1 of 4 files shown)


curran (changes requested):
Awesome start!

cli.js
```diff
@@ -50,11 +53,6 @@ ShareDB.types.register(json1Presence.type);
 
 const app = express();
 
-app.post('/saveTime', (req, res) => {
```
curran, line 53: Tracking this idea in https://github.com/vizhub-core/vzcode/issues/95

cli.js
```diff
@@ -12,6 +12,9 @@ import { randomId } from './src/randomId.js';
 import open from 'open';
 import ngrok from 'ngrok';
 import dotenv from 'dotenv';
+import { diff } from './src/diff.js';
+import chokidar from 'chokidar';
+import { useCallback } from 'react';
```
curran, line 17:
```suggestion
```

cli.js
```diff
@@ -131,26 +129,33 @@ shareDBDoc.subscribe(() => {
...
-        const current = currentDocument[key];
-        if (current.name == oldName) {
-          current.name = filename;
-          shareDBDoc.submitOp([{ p: [key], oi: filename }]);
-          oldName = '';
-          break;
+chokidar.watch(fullPath).on('all', (event, path) => {
+  if (event === 'unlink') {
+    let name = path.split('/').pop();
+    console.log('unlink', name);
```
curran, line 135:
```suggestion
```

cli.js
```diff
@@ -131,26 +129,33 @@ shareDBDoc.subscribe(() => {
...
+    oldName = name;
+  } else if (event === 'add') {
+    if (oldName != '') {
+      let newName = path.split('/').pop();
+      console.log('add', newName);
```
curran, line 140:
```suggestion
```

cli.js
```diff
@@ -131,26 +129,33 @@ shareDBDoc.subscribe(() => {
...
-          oldName = '';
-          break;
+chokidar.watch(fullPath).on('all', (event, path) => {
+  if (event === 'unlink') {
+    let name = path.split('/').pop();
+    console.log('unlink', name);
+    oldName = name;
+  } else if (event === 'add') {
+    if (oldName != '') {
+      let newName = path.split('/').pop();
```
curran, line 139:
```suggestion
      const newName = path.split('/').pop();
```

curran (changes requested):
Very nice! A few minor suggestions.

curran:
It looks like the code is trying to handle renames, but it doesn't work for me on Linux as the `add` event comes first, before the `unlink` event.

Just mapping out here the cases we need to handle:

 * [ ] Rename file
 * [ ] Delete file
 * [ ] Add new file
 * [ ] Text change within file

PR #114, opened 2023-05-31
wip of themes

Anooj-Pai:


src/client/App.jsx
```diff
@@ -46,7 +47,7 @@ function App() {
   const [tabList, setTabList] = useState([]);
 
   // The current theme.
-  const [theme, setTheme] = useState('OneDark');
+  const [theme, setTheme] = useState(oneDark);
```
curran, line 50: That's interesting. While this would work, my gut feeling is to have a string for the state, then use that string to look up the actual plugin implementation.

src/client/getOrCreateEditor.js
```diff
@@ -31,7 +31,7 @@ export const getOrCreateEditor = ({
   // The path for this file in the ShareDB document.
   const path = [fileId, 'text'];
 
-  const extensions = [
+  let extensions = [
```
curran, line 34: I don't see the need for `let` here.
```suggestion
  const extensions = [
```

src/client/getOrCreateEditor.js
```diff
@@ -68,8 +68,7 @@ export const getOrCreateEditor = ({
   }
 
   let editor = editorCache.get(fileId);
-  if (!editor) {
-    // Cache miss --> mint a new editor.
+  if (!editor || editor) {
```
curran, line 71: This doesn't look right. We only want to mint a new editor if there is no existing editor. This is the same as `if(true)`.
```suggestion
  if (!editor) {
```

Maybe let's change it back, and figure out how to swap out the theme plugin on the already created editor. I think there should be a way.

I researched this a bit and it looks like "Compartments" is what we need. See https://codemirror.net/examples/config/#dynamic-configuration

src/client/settings.jsx
```diff
@@ -3,75 +3,28 @@ import Modal from 'react-bootstrap/Modal';
...
-  { value: 'tomorrow-night-eighties', label: 'tomorrow-night-eighties' },
-  { value: 'ttcn', label: 'ttcn' },
-  { value: 'twilight', label: 'twilight' },
-  { value: 'vibrant-ink', label: 'vibrant-ink' },
-  { value: 'xq-dark', label: 'xq-dark' },
-  { value: 'xq-light', label: 'xq-light' },
-  { value: 'yeti', label: 'yeti' },
-  { value: 'yonce', label: 'yonce' },
-  { value: 'zenburn', label: 'zenburn' },
+  { value: abcdef, label: 'abcdef' },
```
curran, line 19: Nice. We can use the `label` strings here in the state.

curran (changes requested):
Great first attempt! I think this is workable. Here's what I see as the next steps:

 * Add the suggested commits from this review
 * Use a string for the state, e.g. `"one-dark"`, rather than the implementation itself. This simplifies the mental model in many ways, and it can be problematic if we pass a function into a `useState` value. For this, we may want to introduce an `id` field to the `themes` array (because `label` may be subject to change over time, and I could imagine in the future storing the user's preferred theme somewhere, and that ID should be stable). Then we can build a `Map` to look up the actual themes based on their id.
 * Introduce a `useEffect` somewhere that changes the theme on all editors in the editor cache using "Compartments" as described in https://codemirror.net/examples/config/#dynamic-configuration

Good luck!

src/client/settings.jsx
```diff
@@ -34,10 +34,12 @@ const saveTimes = [
   { value: 30, label: '30 seconds' },
 ];
 
-export const Settings = ({ show, onClose, setTheme }) => {
+export const Settings = ({ show, onClose, setTheme, editor }) => {
   const handleChange = useCallback((selectedOption) => {
     setTheme(selectedOption.value);
-
+    editor.dispatch({
```
curran, line 40: Putting this here doesn't feel right. Ideally the state change and its side effects should be completely separate. Ideally the `Settings` component should now need to know anything about the `editor`.

I would suggest to put the `editor.dispatch` call into a `useEffect` that depenes on the `theme` state, in a higher level of the component tree.

To implement the theme changing correctly, we may want to consider invoking this on all editors in the editor cache, not just the open editor.

curran (changes requested):
Great progress! I left some feedback for next steps.

curran:
I've been meaning to review this in depth and try it out locally. Hope to do soon!

I'm also looking at swapping in VZCode as the demo editor on this page: https://beta.vizhub.com/

The current state of it uses the default theme, which is not ideal.

curran:
I notice the dropdown itself doesn't show the new selection after you've selected it

image

I can see that the theme does chance, which is amazing! This feels like it's probably pretty close to working.

image

I hope to be able to carve out some time to work on this soon.

PR #126, opened 2023-07-23
Types first attempt

curran:


curran:
This adds `tsconfig` which we may want to do in future, not sure.

Going with the simpler alternative for now https://github.com/vizhub-core/vzcode/pull/125

PR #137, opened 2023-08-23
Prettier

### LINKED ISSUE #71: Auto-Prettier
My ideal code editing experience: [Prettier](https://prettier.io/) should automatically run all the time so I don't need to think about formatting.

Some subtle points:
 * Running prettier _should not_ trigger auto-save, because the semantics of the code will not change
 * How to solve this? Maybe Prettier should trigger with a shorter debounce than auto-save, so that first Prettier will run, then only one auto-save will happen after editing. Cons: this delays auto-save, but maybe that's fine?

Existing Prettier integration in VizHub: https://github.com/vizhub-core/vizhub/blob/a0e06b500b57688082e2262960f816b4286b903f/vizhub-v2/packages/neoFrontend/src/pages/VizPage/PrettierContext/usePrettier.js

curran:
Closes #71


curran:
Closing in favor of https://github.com/vizhub-core/vzcode/pull/140

PR #146, opened 2023-08-27
WIP Display Prettier Errors

### LINKED ISSUE #145: Show Prettier Errors
As a user editing code with VZCode, I want to be able to see the detailed error messages generated by Prettier when it fails.

It would be amazing if they highlighted in the editor, perhaps using https://codemirror.net/docs/ref/#lint.Diagnostic

curran:
Closes #145

Subtasks:
 * [ ] Show error overlay
 * [ ] Highlight errors within CodeMirror editor as red squiggly underline
 * [ ] Add tooltips to red squiggly underline with details of the problem
 * [ ] Add ability to click on a line number in a file in the error stack trace to navigate to that file
 * [ ] Ensure line numbers from sourcemaps are correct

### DIFF
--- a/src/client/usePrettier/index.ts
+++ b/src/client/usePrettier/index.ts
@@ -43,6 +43,8 @@ export const usePrettier = (
     const handleMessage = (event) => {
       const { fileId, error, fileTextPrettified } = event.data;
       if (error) {
+        // TODO display this error here.
+        // See https://github.com/vizhub-core/vzcode/issues/145
         console.log(error);
         return;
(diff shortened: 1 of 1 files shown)


curran:
See also https://github.com/vizhub-core/vizhub/blob/01cadfb78a2611df32f981b1fd8136b70447de9e/vizhub-v2/packages/neoFrontend/src/pages/VizPage/VizRunnerContext/generateRunErrorMessage.js#L4

PR #162, opened 2023-09-03
Split Pane Resizer

curran:
image


curran:
This is what VSCode looks like

image

curran:
Debug view 
image

PR #180, opened 2023-09-21
File delete bug fix

### LINKED ISSUE #40: Delete file

EvWhymark:
Close #40 
This issue was rather difficult to find what was causing it. The attached image highlights which line was creating a problem. What is happening is when you delete a file that has  itself open it was improperly removing itself from the tab list. This was resulting in you searching through the files array for an object that does not exist. There may exist a better solution for this problem however this "hardcoded" approach seems to work for now. Furthermore, I was unable to get a react hook to properly work with my code so if someone can think of the correct syntax for it that would be appreciated(I attempted to use useEffect with the dependencies closeTab, files, tabList however it was improperly calling and the error still occurred).
Screenshot 2023-09-21 023401

### DIFF
--- a/src/client/TabList/index.tsx
+++ b/src/client/TabList/index.tsx
@@ -91,6 +91,13 @@ export const TabList = ({
       );
     };
   }, [handleKeyPress]);
+  //Checks the current state of the tablist and if it is an invalid file it removes it from the files array.
+  for (let i = 0; i < tabList.length; i++) {
+    if (files[tabList[i]] == undefined) {
+      closeTab(tabList[i]);
+      tabList = tabList.splice(i, i);
+    }
+  }
   return (
     <div className="vz-tab-list">
       {files &&


src/client/TabList/index.tsx
```diff
@@ -91,6 +91,13 @@ export const TabList = ({
       );
     };
   }, [handleKeyPress]);
+  //Checks the current state of the tablist and if it is an invalid file it removes it from the files array.
+  for (let i = 0; i < tabList.length; i++) {
+    if (files[tabList[i]] == undefined) {
+      closeTab(tabList[i]);
+      tabList = tabList.splice(i, i);
```
curran, line 98: Thank you for this attempt! My gut feeling is that this is not the ideal place to do this adjustment. The splicing logic is good, but I feel like this should ideally be done elsewhere. This solution here mutates the `tabList` locally, but ideally the _definition_ of `tabList` should change so that the file being deleted is removed from there, then the file could be actually removed from the document and deleted.

The case we are solving for is when you delete a file _and_ that same file is currently open in a tab. Ideally the logic that deletes the file would trickle down in an elegant way to update the tab list.

curran (changes requested):
Thank you for the effort here! The logic is sound, but the placement in the data flow doesn't feel right. Let's try to invoke `closeTab` and mutate `tabList` _before_ removing the file from our ShareDB data structure.

curran:
Closing in favor of #182

DomaDude123:
Looked into the issue and there seems to be a difference between when closeTab is called when removing a tab and deleting a file. When it is called when a tab is removed from the tab list it functions as expected, removing the tab from "tabList". When it is called through "deleteFile" the size of tabList is 0 regardless of how many tabs are open.  Not sure if this is an expected behavior.

PR #182, opened 2023-09-21
Fix file deletion

### LINKED ISSUE #40: Delete file

curran:
Closes #40

Inspired by #180

### DIFF
--- a/src/client/Sidebar/FileListing.tsx
+++ b/src/client/Sidebar/FileListing.tsx
@@ -26,12 +26,12 @@ export const FileListing = ({
     (event) => {
       handleDeleteFileClick(fileId, event);
     },
-    [fileId],
+    [fileId, handleDeleteFileClick],
   );
 
   const handleRenameClick = useCallback(() => {
     handleRenameFileClick(fileId);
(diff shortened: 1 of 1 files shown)


curran:
@EvWhymark Requesting review please. Thanks!

This was suuuuper tricky to figure out. The error turned out to be stemming from the fact that when we call `closeTab` from `deleteFile` in `App.tsx`, the internals of `closeTab` turned out to be using a stale version of the tab list, namely the initial one defined as `[]` empty array. Therefore it was not removing it from the `tabList` before the file actually got deleted.

The fix, interestingly, is to add a missing dependency in the `useCallback` that defines the event listener that ultimately gets added to the button that you click, in `FileListing.tsx`. This change ensures that  we call `closeTab` from `deleteFile` in `App.tsx`, the internals of `closeTab` use the latest version of `tabList`, so that it removes the entry from the `tabList` _before_ the file actually got deleted from the ShareDB document.

Thanks for the inspiration @EvWhymark ! Without your initial push I would not have been able to track this down.

EvWhymark (approved):
Looks good!

PR #183, opened 2023-09-22
Presence: Show Selection

### LINKED ISSUE #131: Presence: Show Selection
Currently only the cursor position of the remote user is shown.

As a collaborative editor, I want to be able to see the selection of the remote user as well.

aruan20:
closes #131 

The goal is to have the user be able to see the selection of the remote user.

### DIFF
--- a/src/client/CodeEditor/json1PresenceDisplay.js
+++ b/src/client/CodeEditor/json1PresenceDisplay.js
@@ -47,15 +47,15 @@ export const json1PresenceDisplay = ({
               const { start, end } = presence;
               const from = start[start.length - 1];
               // TODO support selection ranges (first attempt introduced layout errors)
-              //const to = end[end.length - 1];
+              const to = end[end.length - 1];
               return {
                 from,
-                //to,
-                to: from, // Temporary meaure
+                to,
+                // to: from, // Temporary meaure
                 value: Decoration.widget({
                   side: -1,
                   block: false,
-                  widget: new PresenceWidget(id),
+                  widget: new PresenceWidget(id, from, to),
                 }),
               };
             }),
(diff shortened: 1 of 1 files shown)


aruan20:
Current progress:
The selection range is correct, but it is not being displayed correctly. Instead of highlighting the remote selection, it appears to the user as if the remote user is deleting that part. However, if the remote user clicks out of the selection, the "deleted" part reappears to the user.

Screenshot 2023-09-22 at 2 45 37 AM

curran:
Amazing progress!

Take a look at how they solved it in https://github.com/yjs/y-codemirror/blob/main/src/y-codemirror.js#L160, that might provide some good clues. I'm pretty sure they implemented this exact feature, but based on YJS instead of ShareDB.

curran:
Surfacing this detailed commit comment from @aruan20 :

>I changed the color of the remote selection to be a lighter shade of yellow so that the cursor colors could be visible. However, I'm not sure if the remote cursor should be shown when there is a selection. There are a few options: 1) Not show the cursor and only the selection 2) Show the cursor only at the end (might need to take into account the selection direction because the end could be on the left or right) 3) Show a yellow left border and black right border around the selection. For both option 2 and 3, I solved an issue where adding a border using border-left or border-right shifts the text over by using box-shadow instead. However, when selecting multiple lines, every line shows the border color rather than only at the first and last lines.

A bit of feedback - commenting on the PR or issue is generally better and easier to see. The commit messages get buried in the GitHub UI and can easily be missed.

Digging into the review and local QA now. I'm thrilled to see progress on this!

src/client/CodeEditor/json1PresenceDisplay.js
```diff
@@ -47,17 +47,36 @@ export const json1PresenceDisplay = ({
...
-                //to,
-                to: from, // Temporary meaure
-                value: Decoration.widget({
-                  side: -1,
-                  block: false,
-                  widget: new PresenceWidget(id),
-                }),
-              };
+              const to = end[end.length - 1];
+              if (from == to){
```
curran, line 51:
```suggestion
              if (from === to){
```

curran (approved):
image

Epic!

PR #185, opened 2023-09-22
Add rotate interactive widget

Darmin-Geek:
This allows the user to move their mouse around rotate(x) to change the value of x. This can be used to rotate SVGs (D3). This commit also contains the color picker and vec2 widgets commit. We did not discuss this feature in the meeting, but I thought it was worth exploring after I finished the previous commit.

### DIFF
--- a/src/client/CodeEditor/widgets.ts
+++ b/src/client/CodeEditor/widgets.ts
@@ -104,8 +104,38 @@ export const widgets = ({
           window.open(text);
         },
       },
+
+      //Set rotation to the angle between the x-axis and a line from the word "rotate" to the mouse pointer  (while dragging).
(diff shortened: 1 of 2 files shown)


src/client/CodeEditor/widgets.ts
```diff
@@ -106,9 +106,40 @@ export const widgets = ({
...
+        },
+        onDragEnd(text, setText) {
+          rotationOrigin = null;
+        },
+      },
     ],
   });
 
+
+var rotationOrigin: { x: number; y: number } = null;
```
curran, line 141:
```suggestion
let rotationOrigin: { x: number; y: number } = null;
```

curran (approved):
LGTM

Issue #199, opened 2023-09-23
Usability Improvements: Delete File Confirmation Modal


curran:
As an improvement upon https://github.com/vizhub-core/vzcode/pull/193#pullrequestreview-1640946639, let's introduce a proper modal for confirming file deletion.

curran:
Useful reference: https://react-bootstrap.netlify.app/docs/components/modal/

curran:
Closed by https://github.com/vizhub-core/vzcode/pull/203

PR #203, opened 2023-09-24
WIP Delete Confirmation Modal

DomaDude123:
Implemented the modal confirmation box as was requested in #199 . 

I added a couple of new functions in Item.jsx to handle opening and closing the modal. 

“handleDeleteFileClick” no longer accepts event as an argument as it is no longer called as a result of a button click.

Files referencing "handleDeleteFileClick" were updated to reflect this change. 

Still new to all of this. Please let me know if there is anything that could be improved on or anything I missed. :)

### DIFF
--- a/src/client/Sidebar/Item.jsx
+++ b/src/client/Sidebar/Item.jsx
@@ -1,4 +1,10 @@
-import { useState, useCallback } from 'react';
+import {
+  useState,
+  useCallback,
+  useEffect,
+  useRef,
+} from 'react';
+import { ConfirmationBox } from './confirmationBox';
 
 export const Item = ({
   children,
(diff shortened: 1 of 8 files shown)


src/client/Sidebar/Item.jsx
```diff
@@ -8,6 +14,52 @@ export const Item = ({
...
+
+  //function to handle confirmation on modal
+  const handleConfirmModal = useCallback(() => {
+    setShowModal(false);
+    handleDeleteClick();
+  }, [handleDeleteClick]);
+
+  //function to close Modal if anywhere else on the screen is clicked
+  const outsideClickHandler = useCallback((ref) => {
+    useEffect(() => {
```
curran, line 37: Nope. This is odd. One of the "rules of hooks" is not to invoke hooks inside hooks. They should all be at the same level. See https://react.dev/warnings/invalid-hook-call-warning

In this case we can just remove the useCallback here.

src/client/Sidebar/Item.jsx
```diff
@@ -8,6 +14,52 @@ export const Item = ({
...
+      //check if clicked outside
+      function handleClickOutside(event) {
```
curran, line 39: Interesting... I was under the impression that react-bootstrap handled this interaction of clicking outside the modal to close it.

curran (approved):
Excellent work! I think I'll take over and do a final cleanup pass here, then merge. Thank you!

curran:
@DomaDude123 I cleaned this up and merged. Woohoo!

See https://github.com/vizhub-core/vzcode/pull/203/commits/9efaf89283bdd11a7762fd1ca83164054dc71668 for my changes

Some feedback in general:

 * My preference is to always destructure props rather than use `props.a`, `props.b`, because it makes it easier to know the full set of props when browsing the code.
 * Always good to include links to docs / sources, so others can research later where certain patterns came from

PR #216, opened 2023-09-29
Delete directories

### LINKED ISSUE #102: Directories UX: Delete directory
As a user of VZCode who wants to delete an existing directory, I should be able to do that.

When I delete a directory, all of its children (both files and directories) should be deleted.

EvWhymark:
Closes #102 

Implements the functionality to delete directories and all the children of said directory. Thanks @curran for the advice on how to tackle this problem! I also noticed an issue with FileListing.tsx were the name element was not properly defined. Made this change along with my other commits.

One issue to note with my current logic to performing directory deletion: I am currently passing the path down to App.tsx through the same way we pass down file ids. There is an edge case were this will create problems. If for some reason the user wants to create a directory that is all numbers and is 8 characters long will not allow for directory deletion. We can either make modifications to the logic or prevent users from creating directories that meet these specific properties.


curran (approved):
Notes from review:

 * We should be able to not have to check the IDs with a regex
 * Need to investigate why `name` is being passed into `Item` as both a prop and in the children
 * Multiple tabs open and delete the directory that contains them, there's a crash

PR #220, opened 2023-09-29
Implemented the first version of the hex color picker interactive widget.

Darmin-Geek:


curran (changes requested):
It's almost there! The thing to do now is to investigate the chain of what happens when you call `setText`.

 * [ ] Research to find that definition of `setText` inside the CodeMirror Interact package - somewhere in here https://github.com/replit/codemirror-interact/blob/master/src/interact.ts
 * [ ] Try to isolate the case where `codemirror-ot` fails to translate the CodeMirror change into a ShareDB change.
 * [ ] Ideally we add new tests here: https://github.com/vizhub-core/codemirror-ot/blob/main/test/testTranslation.js

Darmin-Geek:
The issues seem only to apply to the file I was testing on when I demonstrated the feature. Testing other files (even ones with the same contents) show that the feature works. Also, the file I was testing on did not work correctly when edited normally. As a result, I think the issue is not a result of the code I have written. The code in this pull request works.

curran:
Do you know how we could reproduce the error you saw? Having steps to reproduce that would be useful.

PR #221, opened 2023-09-30
Presence: Notifications when user leaves or joins session

### LINKED ISSUE #15: Presence: Notifications when someone joins or leaves the session
Similar in spirit to this UI from VSCode + LiveShare

Image

aruan20:
Closes #15 

Goal: Have a popup at the bottom right that notifies the user when a remote user joins or leaves the collaboration session.


aruan20:
Forgot to comment the status after the last commit. 

Currently, there is a notification after the user leaves the session, which disappears after 10 seconds:
Screenshot 2023-09-30 at 7 04 26 PM

And when there is a someone with a long username, then the user can scroll horizontally to see the rest of the message. But the notification shifts up a bit when the scroll appears, so I might need to fix that so it won't mess with the layout when there are multiple notifications at the same time.

Also, I'm unsure on where to retrieve the usernames to display in the notification, as well as how to check if a user joined the session. If there is a list of usernames, then I could have the notification pop up when the remote user is added to the list, but I'm not sure if there's a list or not.

To summarize, here's a list of things I need to work on:
- add a notification for when the user joins the session
- add usernames to the notifications
- make multiple notifications visible at the same time by having the most recent one at the bottom and pushing the rest up the page

curran (changes requested):
Getting there! Remaining tasks:

 * [x] Accessing the username that the user enters in the settings
 * [ ] This part that listens for presence:
 
```js
         docPresence.on('receive', (id, presence) => {
          // If presence === null, the user has disconnected / exited
          notification(presence);
```
          
          Should not be inside the `src/client/CodeEditor/json1PresenceDisplay.js`, but rather in a higher level of the component tree. Something like this:
          
```js
   // Listen for presence
  useEffect(() => {
    docPresence.on('presence', (id, presence) => {
      ..
    }
  }, [docPresence]);
```

in `App.tsx`.
  
 * [ ] Leverage React to set up the DOM node for the notification.
 * [ ] The CSS can go into a .css file

Suggested approach: make a new component called `PresenceNotifications` that handles display of the notifications.

```jsx
<div className="right">
  <TabList
    files={files}
    tabList={tabList}
    activeFileId={activeFileId}
    setActiveFileId={setActiveFileId}
    closeTab={closeTab}
  />
  {content && activeFileId ? (
    <CodeEditor
      shareDBDoc={shareDBDoc}
      localPresence={localPresence}
      docPresence={docPresence}
      activeFileId={activeFileId}
      theme={theme}
      onInteract={handleInteract}
      editorCache={editorCache}
    />
  ) : null}
  <PresenceNotifications presenceNotifications={presenceNotifications} />
</div>
```

You can follow a similar pattern as `TabList`, where there's an `index` and `styles` file for the DOM manipulation and CSS.

Thanks again for your great work and good luck!

aruan20:
I made a new component PresenceNotifications, as well as added code to retrieve the username from the Settings and send it to the remote user. However, the notifications are not appearing. I tried debugging with the console and it seems like the useEffects in PresenceNotifications doesn't detect that docPresence changes. I want useEffects to be called when the presence changes (because if presence is null, then the user left, etc.), but I'm not sure if docPresence changes when the presence changes.

aruan20:
Commit comment:

> When the remote cursor moves, a notification pops up every time saying that the user joined. A potential solution that I tried to implement is to create an array that keeps track of the users/presence ids that joined the session and then use a conditional statement that checks if the user already joined before displaying a notification. When the user leaves the session, the user would be removed from the array.

src/client/CodeEditor/json1PresenceBroadcast.js
```diff
@@ -1,4 +1,5 @@
 import { EditorView } from 'codemirror';
+import { username } from '../Settings.tsx';
```
curran, line 2: This pattern will not work. We need to pull in the username from some state somewhere, then pass that in.

src/client/Settings.tsx
```diff
@@ -43,6 +45,12 @@ export const Settings = ({
     [],
   );
 
+  const usernameRef = useRef<HTMLInputElement>(null);
+  const handleUsernameChange = () => {
+    const usernameValue = usernameRef.current?.value;
+    username = usernameValue;
```
curran, line 51:
```suggestion
    setUsername(usernameValue);
```

src/client/Settings.tsx
```diff
@@ -14,6 +14,8 @@ const saveTimes = [
   { value: 30, label: '30 seconds' },
 ];
 
+export let username = 'Anonymous';
+
 export const Settings = ({
   show,
   onClose,
```
curran, line 21:
```suggestion
  onClose,
  setUsername,
```

src/client/Settings.tsx
curran: The implementation of `setUsername` will likely feed into the reducer, something like:

```
const setUsername = useCallback((username) => {
  dispatch({type:"set_username", username});
```

curran: Let's focus on this after the main functionality is working.

curran (changes requested):

PR #222, opened 2023-10-03
172 usability improvements tab list goes offscreen

DomaDude123:
Added a very simple, and possibly temporary, fix to the tablist going off screen. When the length of the tab list exceeds the width of the editor a scroll bar appears.


curran (approved):
LGTM!

PR #229, opened 2023-10-10
Test show prettier

nghongwatsang:
WIP


curran:
Remaining tasks:

 * [ ] Make sure the error overlay goes away when Prettier runs successfully with no errors

PR #234, opened 2023-10-13
Delete directory/files bug fixes

### LINKED ISSUE #215: File Contents Stay Open When Deleting Last Tab
If the file you currently have open is the last tab and you go to delete it the files contents will continued to be displayed. You can even continue to edit the file even after deletion resulting in a multitude of errors. View screenshot for example.

image

### LINKED ISSUE #223: Delete Directory With Multiple Tabs Open Crashes
When multiple tabs within a directory are open and then you delete the directory will result in a crash. View attached screenshot.

image

EvWhymark:
Closes #223 
Closes #215 

Decided to bundle the fix to both these bug fixes together since they relate to similar issues. However the solutions to the problems are not related.

The first solution for #223 is related to a pitfall of react's useState. useState as stated in the documentation "does not change the current state in already executing code".  So by calling closeTab multiple times like we used to, resulted in the tabList not being updated at all because of all the conflicting setTabList calls. To solve this issue I created a new method that takes in array of fileIds that we want to close and only uses one setTabList call.

While tackling this issue it made me realize an interesting drawback from using reacts useStates. #223 was difficult to debug and it took a lot of time to carefully track down why our setTabList was incorrectly working. @curran brought up a good point the other day where we could perform a code refractor and change most of or even all of our useStates into a reducer. It would make such cases easier to debug. However, the tradeoff is that useReducer can result in a more complicated code in relation to our useStates. I think this may be worth discussing if we continue to have issues with our useStates as we continue to expand the codebase. See: https://react.dev/learn/extracting-state-logic-into-a-reducer# for more details on reducers.

The solution to #215 interestingly is not related to dependencies, like I originally thought, but was actually an issue with the modal implementation. Previously we had it such that interacting with the sidebar in anyway resulted in the item opening. This meant when you went to delete a file by clicking the trash button resulted in the modal and the file opening. Interestingly if you confirmed the modal/clicked yes this resulted in another event triggering the open of the object which in turn changed the activeFileId again after we already close it in deleteFile(). To solve this it just took a event.stopPropagation() in handleConfirmModal.

While I was solving #215 I also went through the rest of Item.tsx and added more stopPropagations to prevent the file/directory  from opening when you went to delete and rename files.

### DIFF
--- a/src/client/useTabsState.ts
+++ b/src/client/useTabsState.ts
@@ -1,4 +1,4 @@
-import { useCallback } from 'react';
+import { useCallback} from 'react';
 import { FileId } from '../types';
 
 // Encapsulates the state and logic for managing the list of open tabs.
@@ -11,6 +11,7 @@ export const useTabsState = (
   // Called when a tab is closed.
   const closeTab = useCallback(
     (fileIdToRemove: FileId) => {
+      let newActiveFileId: FileId = activeFileId;
       const i = tabList.findIndex(
         (fileId) => fileId === fileIdToRemove,
       );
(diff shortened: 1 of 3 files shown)


src/client/useTabsState.ts
```diff
@@ -29,16 +30,52 @@ export const useTabsState = (
...
+            } else {
+              // or clear out the active file if we've closed the last tab.
+              newActiveFileId = null;
+            }
+          }
+        }
+      });
+      //Run setStates after doing multiple changes to the two variables. Will properly render tabList and activeFileID.
+      setActiveFileId(newActiveFileId);
+      setTabList(newTabList);
```
curran, line 78: Nice! Very elegant how this ends up as just a single call to `setTabList`.

curran (approved):
LGTM! Thank you so much! Excellent work.

I might consider refactoring everything into a `useReducer` at some point, which I think would likely drastically simplify things (but it would be a lot of work).

PR #235, opened 2023-10-13
Refactor setTabList to useReducer

EvWhymark:
Currently correctly refactored the code so setTabList is now used properly within a reducer.

### DIFF
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -3,6 +3,7 @@ import {
   useEffect,
   useCallback,
   useRef,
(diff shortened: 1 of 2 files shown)


curran (approved):
LGTM!

With these large refactorings, I think the best way forward is to actually merge this first increment, then we can migrate things piece by piece, each piece being a separate PR. Otherwise the PR gets stale fast. This approach also allows us to parallelize - I can take a piece, you can take a piece, etc.

Issue #237, opened 2023-10-13
Refactor State to Reducer


curran:
As a developer of VZCode, I want the codebase to be clean and well organized, so that I can easily add new features.

Currently there are many instances of `useState`, and there are operations that need to touch many different state pieces at once. This approach is cumbersome and error-prone. We can refactor these to use `useReducer` to achieve a cleaner codebase with clean separation of concerns across code files.

Pieces of state to migrate into this paradigm (in no particular order):

 * [x] `tabList` #235 
 * [ ] `activeFileId`
 * [ ] `theme`
 * [ ] `isSettingsOpen`
 * [ ] Presence stuff maybe? `localPresence` and `docPresence` Might not make sense... not sure

PR #238, opened 2023-10-13
Refactor useTabState to useReducer

EvWhymark:
#237 

Changes the entirety of useTabState over to our reducer function. I did not delete useTabState.ts however I believe that we can go ahead and remove/archive it if the logic in the reducer function looks good.

### DIFF
--- a/src/client/reducer.ts
+++ b/src/client/reducer.ts
@@ -1,8 +1,99 @@
+import { FileId } from '../types';
+
 export const reducer = (state, action) => {
   switch (action.type) {
     // TODO phase this out
     case 'set_tab_list': {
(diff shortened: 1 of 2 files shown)


src/client/reducer.ts
```diff
@@ -1,23 +1,107 @@
...
+          tabList: [...state.tabList, action.fileId],
+        };
+      } else {
+        return {
+          ...state,
+          activeFileId: action.fileId,
+        };
+      }
+    }
+    case 'close_tab': {
```
curran, line 29: Beautiful!

curran (approved):
LGTM!

PR #239, opened 2023-10-13
Highlight interactive widgets

Darmin-Geek:
User holds alt key to activate the feature.

Note: There is potentially noticeable delay (~2 seconds on a 385 line file with a few places that can use interactive widgets) between pressing the alt key and seeing the highlights. The code I have written checks a regex containing all relevant regexes against every line. The number of checks is likely inevitable. The time it takes to run might not be.

### DIFF
--- a/src/client/CodeEditor/getOrCreateEditor.ts
+++ b/src/client/CodeEditor/getOrCreateEditor.ts
@@ -16,7 +16,11 @@ import {
 } from '../../types';
 import { json1PresenceBroadcast } from './json1PresenceBroadcast';
 import { json1PresenceDisplay } from './json1PresenceDisplay';
-import { colorsInTextPlugin, widgets } from './widgets';
+import {
+  colorsInTextPlugin,
+  highlightWidgets,
+  widgets,
+} from './widgets';
 import {
   EditorCache,
   EditorCacheValue,
@@ -127,7 +131,7 @@ export const getOrCreateEditor = ({
       json1PresenceDisplay({ path: textPath, docPresence }),
     );
 
-    extensions.push(colorsInTextPlugin);
+  extensions.push(colorsInTextPlugin);
 
   // This is the "basic setup" for CodeMirror,
   // which actually adds a ton on functionality.
@@ -185,6 +189,8 @@ export const getOrCreateEditor = ({
   // as they would also need to be able to handle `onInteractEnd`.
   // See https://github.com/replit/codemirror-interact/issues/14
   extensions.push(widgets({ onInteract }));
+
+  extensions.push(highlightWidgets);
 
   const editor = new EditorView({
     state: EditorState.create({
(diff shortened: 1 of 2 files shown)


src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -186,6 +190,8 @@ export const getOrCreateEditor = ({
   // See https://github.com/replit/codemirror-interact/issues/14
   extensions.push(widgets({ onInteract }));
 
+  extensions.push(highlightWidgets);
```
curran, line 193:
```suggestion
  // This extension adds the feature that when the user holds ALT,
  // all possible places where the user can use interactive widgets are highlighted,
  // even if they are not currently hovered over.
  // Note: there is a performance issue with the current version.
  // TODO investigate performance issue
  extensions.push(highlightWidgets);
```

src/client/CodeEditor/widgets.ts
```diff
@@ -312,3 +313,73 @@ class ColorWidget extends WidgetType {
...
+      let offset = 0;
+      while (!line.done) {
+        if (line.value === '\n') {
+          offset++;
+          line = lines.next();
+          continue;
+        }
+        const interactiveOccurances = line.value.matchAll(
+          //The below line contains all of the Regexes for all of the interactive widgets.
+          /\"\#([0-9]|[A-F]|[a-f]){6}\"|rotate\(-?\d*\.?\d*\)|https?:\/\/[^ "]+|vec2\(-?\b\d+\.?\d*\b\s*(,\s*-?\b\d+\.?\d*\b)?\)|true|false|(?<!\#)-?\b\d+\.?\d*\b|rgb\(.*\)/dg,
```
curran, line 346: Very nice! One drawback I see here is that if one of these regexes needs to change for some reason, it needs to be changed here and also in the individual widgets. I wonder if there is a way to pull in the regex definitions from the other modules and combine them here programmatically...

src/client/CodeEditor/widgets.ts
```diff
@@ -312,3 +313,73 @@ class ColorWidget extends WidgetType {
...
+              class: 'cm-interact ',
+            }),
+          };
+        }),
+      );
+    },
+    eventHandlers: {
+      keydown(event, view) {
+        if (event.key == 'Alt') {
+          this.showHighlight = true;
```
curran, line 375: Aha! After setting this, it may be necessary to invoke some method to "update" the CodeMirror editor. I wonder if this might be the 2 second lag - maybe CodeMirror internally has some interval set where it checks for a "dirty" state and updates the display. Perhaps that could be triggered manually, but I'm not sure how.

Perhaps this one? https://codemirror.net/docs/ref/#view.WidgetType.updateDOM

curran (approved):
LGTM!

curran:
Wow this is beautiful!

image

Maybe it might make sense to just show these highlights all the time...

PR #240, opened 2023-10-14
Added missing dependencies. Removed @uiw/codemirror-themes-all

Angra974:
Hi all.

Just made a clean installation and some missing dependencies was found.
Installation was made with Pnpm.

Missing dependencies =>
Capture d’écran du 2023-10-14 19-51-21

Dependencies installation =>
Capture d’écran du 2023-10-14 20-14-33

Not use in the application =>
Capture d’écran du 2023-10-14 20-15-45

Some pictures will be better than a big talk.


curran:
Thank you for this! I'm curious if I am able to reproduce the issue. We have been using `npm` and have never tested with `pnpm`, but it would be great to support both tools.

Indeed, with the themes packages there was some iteration where at first we imported from the `codemirror-themes-all` package, but then changed it to import from the individual packages. This has probably been working fine until now because all of those individual theme packages are dependencies of the `themes-all` package, but maybe `pnpm` is more particular and has uncovered this lack of precision.

I'll do a bit of testing locally with `pnpm` then will likely merge this. Thank you for this contribution!

Angra974:
Pnpm is for lot of people the best package manager. 
You're welcome.

Thanks for your great app, i will use it to build something around it for my personal use first.

curran (approved):
LGTM! Thank you.

PR #242, opened 2023-10-17
WIP Major refactor to use reducer

EvWhymark:


src/client/App.tsx
```diff
@@ -53,26 +46,129 @@ const connection = new Connection(socket);
...
+  const isSettingsOpen = state.isSettingsOpen;
+  const shareDBDoc = state.ShareDBDoc;
+  const setTabList = (tabList) => {
+    dispatch({
+      type: 'set_tab_list',
+      tabList,
+    });
+  };
+
+  const setActiveFileId = (activeFileId) => {
```
curran, line 108:
```suggestion
  const setActiveFileId = useCallback((activeFileId) => {
```

src/client/App.tsx
```diff
@@ -53,26 +46,129 @@ const connection = new Connection(socket);
...
+    dispatch({
+      type: 'set_active_fileId',
+      activeFileId,
+    });
+  };
```
curran, line 113:
```suggestion
  },[dispatch]);
```

src/client/App.tsx
```diff
@@ -53,26 +46,129 @@ const connection = new Connection(socket);
...
+  const openTab = (fileId: FileId) => {
+    dispatch({
+      type: 'open_tab',
+      fileId,
+    });
+  };
+
+  const deleteFile = (fileId) => {
```
curran, line 121: submitOperation((document) => {
        const updatedFiles = { ...document.files };
        delete updatedFiles[action.fileId];
        return { ...document, files: updatedFiles };
      });

src/client/App.tsx
```diff
@@ -53,26 +46,129 @@ const connection = new Connection(socket);
...
+    });
+  };
+
+  // https://react.dev/reference/react/useReducer
+  const [state, dispatch] = useReducer(reducer, {
+    ShareDBDoc: null,
+    tabList: [],
+    activeFileId: null,
+    submitOperation,
+    closeTab,
```
curran, line 82:
```suggestion
```

src/client/reducer.ts
```diff
@@ -95,28 +101,92 @@ export const reducer = (state, action) => {
...
+        shareDBDoc: action.shareDBDoc,
+      };
+    }
+    case 'delete_File': {
+      state.submitOperation((document) => {
+        const updatedFiles = { ...document.files };
+        delete updatedFiles[action.fileId];
+        return { ...document, files: updatedFiles };
+      });
+      state.closeTab(action.fileId);
```
curran, line 127:
```suggestion
```

curran (changes requested):
Looking great! Some suggested changes:

 * Let's make the reducer pure by moving the ShareDB document and the definition of submitOperation outside of the reducer
 * Don't call `dipatch` from inside the reducer, which is what `state.closeTab` does - so remove those methods from the initial state, but do still dispatch the necessary actions, but from the function in `App.tsx`.

curran:
Related to #237 

Closing this as we discussed doing it as a fresh PR that doesn't move the ShareDB logic into the reducer, to keep it "pure".

PR #244, opened 2023-10-18
Rotation indicator for interactive widget

Darmin-Geek:
rotationIndicator1


curran:
Wow very nice! Surprising that it's in the middle of the word, but this is fantastic!

Do you have a sense of how difficult it would be to make it appear to the left or right of the number within parentheses?

curran:
Actually, it makes perfect sense for it to appear where it does, because that's the origin for the interaction!

From the screenshot it looks "wrong", but when you actually interact with it it feels "right".

src/client/CodeEditor/widgets.ts
```diff
@@ -166,22 +166,27 @@ export const widgets = ({
...
         cursor: 'move',
         onDragStart(text, setText, e) {
-          rotationOrigin = { x: e.screenX, y: e.screenY };
+          rotationOrigin = { x: e.clientX, y: e.clientY };
         },
 
         onDrag(text, setText, e) {
           if (rotationOrigin == null) return;
+          const rotationDegree = Math.round(
+            (Math.atan2(
```
curran, line 175: Very nice!

src/client/CodeEditor/widgets.ts
```diff
@@ -383,3 +388,100 @@ export const highlightWidgets = ViewPlugin.fromClass(
...
+
+class RotationCircle extends WidgetType {
+  angle: number;
+  constructor(angle: number) {
+    super();
+    this.angle = angle;
+  }
+
+  toDOM(view: EditorView): HTMLElement {
+    const parent = document.createElement('div');
```
curran, line 442: It looks like this creates a new DOM element each time you move the mouse. It's not a big deal as performance seems fine, but something I wonder if we could optimize by re-using the existing DOM element and just changing the indicator line.

curran (approved):
Excellent work!

PR #247, opened 2023-10-20
WIP Presence Notifications Improved

aruan20:
Goal: Improve presence notifications

- [x] fix timing issue of when notifications leave the screen (should leave one by one from the top and not clumps)
- [x] users that join the session should not be notified about users that have already joined the session
- [x] usernames (might wait until after the reducer is merged)


aruan20:
The timer function is changed so that instead of filtering the array of presence notifications for notifications that don't have the presence ID, it removes the first notification from the array, which is the notification that was displayed first. 

https://github.com/vizhub-core/vzcode/assets/90737103/419026d5-7680-43cb-8240-0c615faf2f55

curran:
Very nice!

I'm beginning to wonder if maybe we need to take another approach to _tracking_ the presence of users in a session. The ShareDB presence API seems quite limited in that there seems to be no way of knowing who is already in the session when you join.

I dropped a note about this in the ShareDB repo: https://github.com/share/sharedb/issues/632

curran:
@aruan20 Check out this thread in ShareDB, which may contain a solution to our problem: https://github.com/share/sharedb/issues/632

aruan20:
I implemented the username function by adding to the reducer and passing the username to the local presence as a variable. 

At first, I tried passing the username into the CodeEditor, but this caused the username to only update every time the user opens a new file. If the user opened a file and changed the username, the username doesn't get updated until the user opens a different file.

The current method – passing the username as a variable in local presence – works in most cases, except for when the user enters a file, changes their username and leaves without changing their local presence (move their cursor, etc.). However, if they move their local presence, their username will update even if they don't open a different file.

Currently:

https://github.com/vizhub-core/vzcode/assets/90737103/0a4a817a-574c-4c94-b191-8dfcfee7f0fd

https://github.com/vizhub-core/vzcode/assets/90737103/79d2f90c-4fae-4be4-9e18-d7a052f1864b

Case where it doesn't work:

https://github.com/vizhub-core/vzcode/assets/90737103/bd903b75-b88a-4311-8eb0-e47333d5b618

aruan20:
Thanks @curran! The custom presence ids helped solve the issue.

I passed the local presence into the PresenceNotifications component which compares the timestamp of the remote user's presence id to the local user's presence id. If the remote user joined before the local user, the code returns without displaying a notification.

https://github.com/vizhub-core/vzcode/assets/90737103/fccc3510-bfcc-44c1-bf87-d927538e86b6

curran:
Amazing! Let's go over this and resolve merge conflicts together on our call.

curran (approved):
LGTM!

PR #249, opened 2023-10-20
WIP: Test prettier error overlay resize

nghongwatsang:


curran:
Thanks! But what issue does this close? What feature does this add?

See https://github.com/vizhub-core/vzcode/blob/main/CONTRIBUTING.md#pull-requests

>No Associated Issue: If you're enthusiastic about something that doesn't have an associated issue, that's great! But make sure you create an issue first, mainly to validate your idea before spending time on a PR. This ensures we track all changes and discussions.

This is what I see when I test this branch:

image

image

image

I'm not sure what the direction of the work is here. Please clarify what you are trying to accomplish. Thanks!

nghongwatsang:
I was experimenting with the CSS and then noticed that the area where errors are typically displayed lacked the functionality to open and close, whether by dragging or pressing a button. Initially, I opted to implement the dragging feature to facilitate opening and closing. However, I soon realized that creating a button for this purpose would be a more suitable approach for our setup.

The concept was to have prettier errors displayed below but grant the editor the ability to close the popup temporarily and reopen it as needed. I made the decision to switch to this approach just this morning, which is why it remains incomplete.

curran:
>lacked the functionality to open and close

Nice observation! I would recommend to start a discussion about whether the functionality of opening and closing the prettier error overlay is needed at all, before starting work on an implementation.

I'm curious, why do you think this functionality would be useful? In what scenario would one want to close the error overlay?

nghongwatsang:
My thought was that it could be an additional feature for those who don't want something to constantly pop up at the bottom of the screen when there's an error. For those individuals, a more subtle option could be implemented to indicate that a syntax error has occurred.

But, now that I'm thinking about it, this could be connected to the underlining of syntax error issues we already have. Instead of our current way of displaying errors, the first step would be to underline the syntax errors. Once a user clicks on the red squiggly line, the display below would pop up. The open/close button would potentially add further functionality to this system.

In addition, even though the error display box is small, some users might want to shrink it to get a full view of their code.

curran:
Awesome! Now we're getting somewhere.

I like to frame changes as "user stories", which clarifies the pain point and also opens up lots of possibilities for solutions.

Check this out https://en.wikipedia.org/wiki/User_story

>As a <role> I want to <capability>, so that <receive benefit>

I get the issue that you see - summarized as

 * "don't want something to constantly pop up"
 * "shrink it to get a full view of their code."

Maybe we can frame it like this:

"As a user editing code that has a syntax error, I want to be able to edit the code that is beneath the error overlay"

Immediately this opens up an alternative possibility, which is to not have it as an overlay, but actually have it occupy the bottom portion of the display, which would let the user _scroll_ within the editor, being able to edit all the code again.

Modifying it in that way would address the issue of not being able to edit code beneath the overlay, and also would be a relatively light lift (just style / layout changes).

curran:
I made a new issue for this @nghongwatsang https://github.com/vizhub-core/vzcode/issues/250

curran:
I'm closing this, as I think it makes sense to start fresh towards #250

PR #261, opened 2023-10-24
WIP Implement intelligent autocompletion

EvWhymark:
This is a very early prototype and has a bunch of issues. Thought I would create a PR to log what I have found and the important notes that I have noticed while working on this implementation. 

Important notes about my implementation:

- If you want to get this running locally just uncomment the extensions.push on line 202.
- I was able to get the typescript server running in the browser by using typescripts virtual file system [@typescript/vfs](https://github.com/microsoft/TypeScript-Website/tree/v2/packages/typescript-vfs). This approach feels far simpler and easier to manage than hosting our our TypeScript server in a web worker. I was able to get the web worker working in my testing however it was far more complicated than this alternative approach.  If for some reason in the future we need the web worker approach it is definitely doable. 
- I was unable to get the code to properly load any file that isn't a typescript file. I enabled enableJs in the tsconfig.json file and in compilation options for both of our environments however this didn't resolve the error. 
- On the topic of files, the current typescript environment that I am setting up does not accept the multi editor approach we are using for multiple tabs. The issue is on lines 220 and 227. This is a pretty major issue however I was more focused on working with the autocompletions with this commit. If we want to continue with this approach we will need to modify the root files to be in line with our editors file system.
- The corrections work decently currently when you are inside a typescript file. It can properly list recommendations and will filter them to match with what you are typing. The issues start to arise when you start to go into methods of objects. For example when typing console.log the code will correctly autocomplete console however the .log part confuses the typescript server and starts recommending incorrect methods for console. 
- Currently we are auto running prettier which I think is a fine idea however we can not have it on such a small interval if we want there to be intelligent autocompletions. It interferes with the users selection on the autocompletions.
- The way codemirrors keymappings are set up is very interesting. Typically when you are autocompleting text within editors you will use tab to autcomplete however codeMirror believes this to be too inaccessible.  We will need to manually modify the autocompletions keymapping. Interesting read on the topic: https://codemirror.net/examples/tab/

Some important resources I used in my exploration:

- Code took heavy inspiration from: https://stackblitz.com/edit/codemirror-6-typescript?file=package.json
- Documentation for typescript vfs: https://github.com/microsoft/TypeScript-Website/tree/v2/packages/typescript-vfs
- Interesting post about implementing this feature into code mirror: https://discuss.codemirror.net/t/codemirror-6-and-typescript-lsp/3398/3
- Another example using the webworker approach (Code is really difficult to read and I believe not well written): https://codesandbox.io/s/typescript-lsp-c3mqx?file=/public/workers/tsserver.ts
- Codemirror autocompletion example: https://codemirror.net/examples/autocompletion/
- Codemirror autocompletion documentation: https://codemirror.net/docs/ref/#autocomplete

### DIFF
--- a/src/client/CodeEditor/getOrCreateEditor.ts
+++ b/src/client/CodeEditor/getOrCreateEditor.ts
@@ -27,6 +27,9 @@ import {
   EditorCacheValue,
 } from '../useEditorCache';
 import { ThemeLabel, themeOptionsByLabel } from '../themes';
+import * as tsvfs from '@typescript/vfs';
+import ts from 'typescript';
+import { autocompletion } from '@codemirror/autocomplete';
 
 // Language extensions for CodeMirror.
 // Keys are file extensions.
@@ -56,15 +59,15 @@ const getAtPath = (obj, path) => {
 // Gets or creates an `editorCache` entry for the given file id.
 // Looks in `editorCache` first, and if not found, creates a new editor.
 export const getOrCreateEditor = ({
-  fileId,
-  shareDBDoc,
-  filesPath,
-  localPresence,
-  docPresence,
-  theme,
-  onInteract,
-  editorCache,
-}: {
+                                          fileId,
+                                          shareDBDoc,
+                                          filesPath,
+                                          localPresence,
+                                          docPresence,
+                                          theme,
+                                          onInteract,
+                                          editorCache,
+                                        }: {
   fileId: FileId;
 
   // The ShareDB document that contains the file.
@@ -194,6 +197,69 @@ export const getOrCreateEditor = ({
   extensions.push(highlightWidgets);
 
   extensions.push(rotationIndicator);
+
+/*
+  extensions.push(autocompletion({
+    override: [tsComplete]
+  }))*/
+
+  // displays autocompletes
+  async function tsComplete(ctx) {
+
+    const compilerOptions = {};
+    const fsMap = await tsvfs.createDefaultMapFromCDN(
+      compilerOptions,
+      ts.version,
+      true,
+      ts
+    );
+    fsMap.set(name, text);
+    const system = tsvfs.createSystem(fsMap);
+    const env = tsvfs.createVirtualTypeScriptEnvironment(
+      system,
+      [name],
+      ts,
+      {
+      }
+    );
+
+    let tsCompletions = env.languageService.getCompletionsAtPosition(
+      name,
+      ctx.pos,
+      {}
+    );
+
+    if (!tsCompletions) return { from: ctx.pos, options: [] };
+
+    const Text = ctx.state.doc.toString();
+
+    let lastWord, from;
+    for (let i = ctx.pos - 1; i >= 0; i--) {
+      if ([' ', '.', '\n', ':', '{'].includes(Text[i]) || i === 0) {
+        from = i === 0 ? i : i + 1;
+        lastWord = Text.slice(from, ctx.pos).trim();
+        break;
+      }
+    }
+
+    if (lastWord) {
+      tsCompletions.entries = tsCompletions.entries.filter((completion) =>
+        completion.name.startsWith(lastWord)
+      );
+    }
+
+    return {
+      from: ctx.pos, // Autocomplete position
+      options: tsCompletions.entries.map((completion) => ({
+        label: completion.name,
+        apply: (view) => {
+          view.dispatch({
+            changes: { from, to: ctx.pos, insert: completion.name },
+          });
+        },
+      })),
+    };
+  }
 
   const editor = new EditorView({
     state: EditorState.create({


src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -195,6 +198,69 @@ export const getOrCreateEditor = ({
...
+/*
+  extensions.push(autocompletion({
+    override: [tsComplete]
+  }))*/
+
+  // displays autocompletes
+  async function tsComplete(ctx) {
+
+    const compilerOptions = {};
+    const fsMap = await tsvfs.createDefaultMapFromCDN(
```
curran, line 210: This approach creates one `fsMap` for each editor. We want to only ever create one single one of these.

I think the CodeMirror extension should just get the completions by passing messages into a Web Worker that contains a singleton `fsMap`.

There will be a whole part of this system that lives outside the CodeMirror extension, namely:

 * Something that sets up the Web Worker
 * Something that keeps the `fsMap` within the Web Worker in sync with our ShareDB-based file system

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -195,6 +198,69 @@ export const getOrCreateEditor = ({
...
+      compilerOptions,
+      ts.version,
+      true,
+      ts
+    );
+    fsMap.set(name, text);
```
curran, line 216: This is cool, but we need to add the content from _all_ our files to this VFS.

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -195,6 +198,69 @@ export const getOrCreateEditor = ({
...
+    const system = tsvfs.createSystem(fsMap);
+    const env = tsvfs.createVirtualTypeScriptEnvironment(
+      system,
+      [name],
+      ts,
+      {
+      }
+    );
+
+    let tsCompletions = env.languageService.getCompletionsAtPosition(
```
curran, line 226: This call should happen inside the Web Worker, then the completions should be serialized and sent as a message back to the CodeMirror extension.

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -195,6 +198,69 @@ export const getOrCreateEditor = ({
...
+      {}
+    );
+
+    if (!tsCompletions) return { from: ctx.pos, options: [] };
+
+    const Text = ctx.state.doc.toString();
+
+    let lastWord, from;
+    for (let i = ctx.pos - 1; i >= 0; i--) {
+      if ([' ', '.', '\n', ':', '{'].includes(Text[i]) || i === 0) {
```
curran, line 238: It's worth adding a comment here describing what this cryptic code is doing. If it's copy-pasted from somewhere, best to include the URL to the upstream code in a comment.

curran (changes requested):
Excellent progress! I've dropped a bunch of comments. I'll synthesize here what I think are the most important points:

Here's what I envision as the path forward for the architecture:

 * We only ever have _one_ instance of the TypeScript virtual file system
 * That instance lives inside a Web Worker
 * We throttle updates from our ShareDB-based data structure for files, so that at most once every second or so we update the content of files in the VFS (updating _only_ the ones that changed). This is similar to what the VZCode server is already doing to keep the actual file system in sync with edits.
 * Each CodeMirror editor instance has a very lightweight extension that just communicates with the singleton VFS through the WebWorker message passing API. In other words, the CodeMirror extension should just get the completions by passing messages into a Web Worker that contains a singleton `fsMap`, and listening for the messages that come back out.

There will be a whole part of this system that lives outside the CodeMirror extension, namely:

 * Something that sets up the Web Worker
 * Something that keeps the `fsMap` within the Web Worker in sync with our ShareDB-based file system
 
 Here's what I would recommend as the next steps:
 
  * [ ] Set up the Web Worker. Let's bite the bullet and get this done, as it will ultimately be the best move, and will set us up with a "central brain" that we can send messages to and receive messages from
  * [ ] Set up the part of the system that keeps the VFS in sync with our ShareDB document. This can be a custom hook used in `App.tsx`, something like `useTypeScriptVFSSync`. This would be responsible for passing the ShareDB files into the Web Worker, maybe throttled/debounced. When the Web Worker receives this message, it can update the files in the VFS to be in sync. As a first pass, we can provide all the files and update all the files. As a later optimization, we can just pass in the files that changed.
  * [ ] Modify the CodeMirror extension to get the completions by passing a message into the Web Worker, then receiving the corresponding message back out. Keep in mind that every extension instance will receive all messages that the Web Worker emits, so we may need to come up with some scheme to figure out which messages are for which editors. Perhaps we can leverage `fileId` for this.

Great start, and good luck!

curran:
Our ultimate goal: Get these to complete correctly for D3 methods:

image

curran:
Try this from GPT4:

```js
import * as tsvfs from '@typescript/vfs';
import ts from 'typescript';

// Assuming you have compilerOptions defined somewhere above
const fsMap = await tsvfs.createDefaultMapFromCDN(
  compilerOptions,
  ts.version,
  true,
  ts
);

const system = tsvfs.createSystem(fsMap);

// Note: 'env' is only created for the sake of demonstrating its creation.
// For this modification, we don't directly use it. However, it can be
// useful in other parts of your project where you might need a virtual
// TypeScript environment.
const env = tsvfs.createVirtualTypeScriptEnvironment(
  system,
  [], // the array of root files can be dynamically set based on the scenario.
  ts,
  {}
);

// Saves the files that changed.
const save = () => {
  const currentDocument = shareDBDoc.data;

  // Take a look at each file (key) previously and currently.
  const allKeys = Object.keys({
    ...currentDocument.files,
    ...previousDocument.files,
  });
  for (const key of allKeys) {
    const previous = previousDocument.files[key];
    const current = currentDocument.files[key];

    // If this file was neither created nor deleted...
    if (previous && current) {
      // Handle changing of text content.
      if (previous.text !== current.text) {
        system.writeFile(current.name, current.text);
      }

      // Handle renaming files.
      if (previous.name !== current.name) {
        const content = system.readFile(previous.name);
        if (content) {
          system.deleteFile(previous.name);
          system.writeFile(current.name, content);
        }
      }
    }

    // Handle deleting files and directories.
    // Note: The TypeScript VFS does not differentiate between directories and files.
    // So, we'll simply remove the file using the system.
    if (previous && !current) {
      system.deleteFile(previous.name);
    }

    // Handle creating files.
    if (!previous && current) {
      system.writeFile(current.name, current.text);
    }
  }

  previousDocument = currentDocument;
};

```

curran:
For moving forward, I would recommend:

 * [ ] Create a brand new PR that _only_ sets up the Web Worker, sends a test message to it to populate the VFS, generates an example completion, and logs the result

From there we can map out the subsequent steps as separate issues that can each be tackled in separate PRs.

curran:
From our discussion:

 * There is a hard barrier between the existing app and the Web Worker
 * All we can do from the App, which has ShareDB access, is to send messages to the worker and receive messages from the worker
 * The Web Worker will have access to the TypeScript VFS, but _not_ ShareDB directly
 * As a first step, let's send a message to the Web Worker that has the entire content of the `content` variable, which contains all the files we track
 * In the Web Worker, it can receive that message and populate the VFS
 * Next, the CodeMirror extension should send a message to the Web Worker to request a completion
 * The Web Worker receives that message, generates the completion, and passes it back
 * The CodeMirror extension then receives that message and presents the completion

Issue #103, opened 2023-05-24
Rename directory


curran:
As a user of VZCode, I want to be able to rename a directory.

Implementation detail: all the children of the renamed directory will need to be updated.

Related to
 * #105
 * #150 
 * #173

curran:
Start in `src/client/Sidebar/DirectoryListing.tsx` line 42

curran:
Similar to work done in https://github.com/vizhub-core/vzcode/pull/216

curran:
Closed in https://github.com/vizhub-core/vzcode/pull/567

PR #264, opened 2023-10-24
Refactor deletion

### LINKED ISSUE #224: Delete Directory Confirmation Modal
When a user deletes a directory it  prompts them with a confirmation as if they were deleting a file.

### LINKED ISSUE #263: Refactor File & Directory Deletion
Currently there is a "bad smell" in our CRUD, namely:

```
  // Generic delete function that calls either deleteFile or deleteDirectory
  // TODO consider if there is a cleaner way to do this
  //  - Ideally we would have a separate function for deleting files and directories
  //  - When we click the delete button, we should be able to tell if it is a file or directory
  const handleDeleteClick = useCallback(
    (key: string) => {
      // Regex to identify if the key is a file path or a file id.
      if (/^[0-9]*$/.test(key)) {
        if (key.length == 8) {
          deleteFile(key);
        } else {
          deleteDirectory(key);
        }
      } else {
        deleteDirectory(key);
      }
    },
    [deleteFile, deleteDirectory],
  );
```

This feels wrong and I think it could lead potentially to unexpected bugs later on. The logic is very brittle. Let's fix it.

curran:
Closes #224. Closes #263.

Changed UI and copy to mimic that of OpenAI:

image

Handle directories in particular:

image

Removed the rename button for directories, put it behind a feature flag for when we work on #103 

image

Also added types here and there, and made the rename input auto-focus.


EvWhymark (approved):
Looks good to me!

PR #265, opened 2023-10-24
New file shortcut - WIP

DomaDude123:
**Added functionality:**
Added a keyboard shortcut allowing users to create a new file with alt-n. The actual file creation process remains the same. When a new file is created it is set as active in the editor window. 

**Future functionality**
The new file is opened in the editor window but it requires users to manually click on the window in order to start typing.
The file naming prompt seems to be the browser default. Maybe we could implement a nice looking UI as we've done for deleting a file?


src/client/useFileCRUD.ts
```diff
@@ -7,20 +7,23 @@ import { randomId } from '../randomId';
...
   closeTabs,
+  openTab,
 }) => {
   // Create a new file
   const createFile = useCallback(() => {
     // TODO use Bootstrap modals not native prompts
     // See https://github.com/vizhub-core/vzcode/issues/252
     const name = prompt('Enter new file name');
     if (name) {
+      let idn = randomId()
```
curran, line 18:
```suggestion
      const newFileId: FileId = randomId()
```

curran (approved):
Excellent!

curran:
I had a new insight on how we can deal with the editor focus and avoid the `setTimeout` hack:

 * When we know that an editor should be focused, we queue it up to be focused when the editor renders by setting some state like `editorWantsFocus: true` in the reducer state.
 * Inside `CodeEditor`, which has access to the editor cache and the editor that needs focus, we can check `editorWantsFocus`, and if it's `true`, we focus the editor and set it to false.

This would avoid the `setTimeout` hack, and also would make the logic of `focusEditor` simpler, as it would not need to access the editor cache directly.

curran:
>The file naming prompt seems to be the browser default. Maybe we could implement a nice looking UI as we've done for deleting a file?

Agreed! Created a new issue for that: #266

PR #284, opened 2023-10-27
WIP: Test show and remove prettier errors 

nghongwatsang:
Issue: 250
I ended up adding some additional css to the x button. In addition, when I removed the overlay, I was having trouble bringing the overlay back. So I just added a button that would allow the overlay to reappear if pressed. I don't know if you like this implementation better, but I wanted to show another possible option before doing it the initial way it was intended.


src/client/CodeErrorOverlay/index.tsx
```diff
@@ -1,27 +1,37 @@
...
-// Feature flag for the button that dismisses the error.
-const enableErrorDismiss = false;
+const enableErrorDismiss = true;
 
 export const CodeErrorOverlay = ({
   errorMessage,
 }: {
   errorMessage: string | null;
 }) => {
+  const [isOverlayVisible, setIsOverlayVisible] = useState(true);
```
curran, line 11: Here's one way we can reset the dismissal every time the error message is updated.

```suggestion
  const [isOverlayVisible, setIsOverlayVisible] = useState(true);
  
  // If errorMessage changes, set the overlay to be visible
  useEffect(() => {
    if (errorMessage !== null) {
      setIsOverlayVisible(true);
    }
  }, [errorMessage]);
```

curran (changes requested):
Very nice!

Instead of having a new button that resets the dismissal (calls `setIsOverlayVisible(true)`), we can reset the dismissal automatically every time the error changes. I've attached a suggested implementation that uses `useEffect`.

Issue #290, opened 2023-10-28
AI Assist - Handle Simultaneous Edits


curran:
As a user who may be editing text that comes before the place where AI is inserting text during generation, I want the output to be correct and not crash with server errors.

To reproduce this error:

```
Editor is live at http://localhost:3030
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: The op is too long for this document
    at Object.skip (/home/curran/repos/vzcode/node_modules/sharedb-client-browser/dist/ot-json1-presence-umd.cjs:2807:27)
```

Do the following:

 * Delete a character that comes before the place where the AI is generating, before it has finished generating

curran:
Some ideas for handling simultaneous edits:

```js
shareDBDoc.on('op', (ops, source) => {
  if (source !== "AIAssist") { // Assuming this differentiates local from remote ops - this logic may not be right
    ops.forEach(op => {
      insertionCursor = type.transformPosition(insertionCursor, op);
    });
  }
});

```

See also https://github.com/ottypes/text-unicode#transforming-cursor-positions

There may be addtional considerations like:

 * We only want to consider the ops that impact this file - see

curran:
This may also be useful - the logic that's used for transforming presence:

https://github.com/vizhub-core/json1-presence/blob/master/lib/json1.ts#L733

```js
function transformPresence(presence: Presence | null, op: JSONOp, isOwnOperation: boolean): Presence | null {
    if (!presence || !presence.start || !presence.end) {
        return null
    }

    const start = transformPosition(presence.start, op)
    const end = transformPosition(presence.end, op)

    if (start && end) {
        return { ...presence, start, end }
    } else if (start) {
        return { ...presence, start, end: start }
    } else if (end) {
        return { ...presence, start: end, end }
    }

    return null
}
```

In our case, it would use `insertionCursor` instead of  `presence.start` and `presence.end`.

PR #291, opened 2023-10-28
Reliably reproduce op error

curran:
Towards #290 

@Darmin-Geek This can be a good starting point - a way to easily reproduce the op error.


curran:
Closing as this is dated

PR #293, opened 2023-10-30
Fix Cursor Bug

### LINKED ISSUE #206: Presence: Fix multiple cursors bug
Current behavior: the cursor of a remote editor displays _for every line_ when multiple lines are selected.

TODO drop screenshot

Desired behavior: either of the following, whichever is easier:
 * Don't show the cursor at all when there is a selection (not optimal but perhaps easiest)
 * Only show the actual remote cursor position once, even when multiple lines are selected (it's possible this may not be feasible, if the underlying presence representation does not express the "one true cursor" - maybe it's just a list of selections and there's no way of knowing where the actual cursor is - unless we make a change to how presence is represented)

aruan20:
Goal: Fix issue where a multiple line selection causes a cursor to appear at the end of each line

Closes #206

### DIFF
--- a/src/client/CodeEditor/json1PresenceDisplay.ts
+++ b/src/client/CodeEditor/json1PresenceDisplay.ts
@@ -52,6 +52,7 @@ export const json1PresenceDisplay = ({
               const from = start[start.length - 1];
               // TODO support selection ranges (first attempt introduced layout errors)
               const to = end[end.length - 1];
+              
               if (from === to) {
                 return {
                   from,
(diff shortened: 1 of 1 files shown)


aruan20:
The new commit doesn't show the cursor when there is a selection. Another solution would be to only show the cursor on the last selected line, but I wasn't able to figure out how to have the cursor appear at a specific location.
Screenshot 2023-10-30 at 5 58 48 PM

src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -70,8 +71,8 @@ export const json1PresenceDisplay = ({
                     class: 'cm-json1-presence',
                     attributes: {
                       style: `
-                        background-color: rgba(255, 255, 0, 0.5);
-                        box-shadow: 1px 0 0 yellow;
+                        background-color: rgba(255, 255, 0, 0.3);
+                        box-shadow: 1px 0 0 clear;
```
curran, line 75:
```suggestion
```

If there's no box shadow, I think we could just remove this line.

curran (approved):

curran (approved):
LGTM!

PR #294, opened 2023-10-31
Intelligent autocompletion

### LINKED ISSUE #147: Provide Intelligent Autocompletions
Subtasks:

TODO split these into separate issues

 * Run the TypeScript language server in a Web Worker
 * Leverage TypeScript language server for CodeMirror completions
 * Suggest completions to local files when importing
 * Suggest completions of variable names based content of local files

EvWhymark:
Closes #147 

Finally implemented the functionality to perform intelligent autocompletions! Thanks for the help on completing this implementation @curran! If you want to get this running locally just change the feature flag within getOrCreateEditor.ts to true and make sure you run it with npm run dev (otherwise the web worker will not work). This implementation uses a [shared web worker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker) and performs all typescript server operations inside it. This is the first pass on the implementation and can be heavily improved upon but is a good start.

Some features/issues that need to be addressed in the future:
- Currently only takes in typescript files. It should also take in .js files however I was struggling to get the typescript compiler to accept this.
- The code is poorly optimized. The most notable optimization to make is within App.tsx. Currently we are updating the content of the typescript file system every time we make a change in any file. What we want to happen is that it only updates when the file system changes (file is created/deleted/renamed/moved). 
- I am realizing upon writing this PR that I did not properly modify the typescript file system to account for directories. I have not tested it however, any typescript file, in theory, will not work if it is in a directory.

There are bound to be more issues however these are just the ones that I noticed while writing my implementation. 

I also commented out the AIAssist in the getOrCreateEditor because my local version does not have the API key and was creating too many errors for me. We can uncomment it out before merging if we want.

### DIFF
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -74,6 +74,21 @@ function App() {
   const [content, setContent] =
     useState<VZCodeContent | null>(null);
 
+  let tsServer = new SharedWorker(
+    new URL('src/worker.js', window.location.origin),
+    {
+      name: 'worker',
+      type: 'module',
+    },
+  );
+
+  //TODO: Make this only happen on file changes
+  tsServer.port.start();
+  tsServer.port.postMessage({
+    event: 'update-content',
+    details: content,
+  });
+
   // Set up the connection to ShareDB.
   // TODO move this logic to a hook called `useShareDB`
   useEffect(() => {
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -74,7 +74,7 @@ function App() {
   const [content, setContent] =
     useState<VZCodeContent | null>(null);
 
-  let tsServer = new SharedWorker(
+  const tsServer = new SharedWorker(
     new URL('src/worker.js', window.location.origin),
     {
       name: 'worker',
--- a/src/client/CodeEditor/getOrCreateEditor.ts
+++ b/src/client/CodeEditor/getOrCreateEditor.ts
@@ -261,7 +261,7 @@ export const getOrCreateEditor = ({
 
     //Logic to get the text and cursor location in between punctuation.
     //Taken from https://codemirror.net/examples/autocompletion/
-    let from = ctx.matchBefore(/\w*/).from;
+    const from = ctx.matchBefore(/\w*/).from;
     let lastWord = ctx.matchBefore(/\w*/).text;
     if (lastWord) {
       // @ts-ignore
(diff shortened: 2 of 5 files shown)


curran:
Remaining TODOs:

 * [ ] only takes in typescript files. It should also take in .js files
 * [ ] Optimization - Currently we are updating the content of the typescript file system every time we make a change in any file. What we want to happen is that it only updates when the file system changes (file is created/deleted/renamed/moved).
 * [ ] account for directories
 * [ ] AIAssist should not crash if developer does not have OpenAI key

src/client/App.tsx
```diff
@@ -74,6 +74,21 @@ function App() {
   const [content, setContent] =
     useState<VZCodeContent | null>(null);
 
+  let tsServer = new SharedWorker(
```
curran, line 77:
```suggestion
  const tsServer = new SharedWorker(
```

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -29,7 +29,9 @@ import {
 } from '../useEditorCache';
 import { ThemeLabel, themeOptionsByLabel } from '../themes';
 import { AIAssist } from '../AIAssist';
+import { autocompletion } from '@codemirror/autocomplete';
 
+let useAutoComplete = false;
```
curran, line 34:
```suggestion
const useAutoComplete = false;
```

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -207,14 +209,87 @@ export const getOrCreateEditor = ({
...
+  if (useAutoComplete) {
+    extensions.push(
+      autocompletion({
+        override: [tsComplete],
+      }),
+    );
+  }
+
+  //Creating SharedWorker. Used in autocompletions
+  let tsServer = new SharedWorker(
```
curran, line 230:
```suggestion
  const tsServer = new SharedWorker(
```

src/worker.js
```diff
@@ -0,0 +1,117 @@
+import * as tsvfs from '@typescript/vfs';
+import ts from 'typescript';
+
+let previousDocument = null;
+let currentDocument = null;
+
+let compilerOptions = {};
```
curran, line 7:
```suggestion
const compilerOptions = {};
```

curran (approved):

PR #298, opened 2023-10-31
Prevent crash when deleting text

Darmin-Geek:
Bug: making changes to the document while the AI Assist is running leads to the AI cursor moving to the 0th index.


curran (approved):
LGTM! Amazing work.

PR #305, opened 2023-11-02
Intelligent Autocompletions

curran:
Builds out the foundations for 

 * #147 

Done in an epic pair programming session with @EvWhymark , based on previous work found in https://github.com/vizhub-core/vzcode/pull/294 and https://github.com/vizhub-core/vzcode/pull/295.


src/client/useTypeScript/index.ts
```diff
@@ -0,0 +1,86 @@
...
+// }    ,[content])
+
+// // // Updates the TypeScript worker with the current content
+// // // of the files.
+// // typeScriptWorker.postMessage({
+// //     event: 'update-content',
+// //     details: content,
+// //   });
+
+// };
```
EvWhymark, line 86:
```suggestion
```
Should probably remove unnecessary comment.

src/client/useTypeScript/index.ts
```diff
@@ -0,0 +1,86 @@
+//     import { JSONOp } from 'ot-json1';
+// import {
+//   FileId,
+//   ShareDBDoc,
+//   VZCodeContent,
+// } from '../../types';
+// import { useEffect, useRef, useState,useCallback } from 'react';
```
EvWhymark, line 7:
```suggestion
```
Unnecessary comment

EvWhymark (approved):
Just some leftover comments from our thinking that we should probably remove but otherwise looks good to me! Really glad we were able to finish this up and get it working.

curran:
It "works" but the completions are generic, something is wrong.

curran:
Got it!

image

PR #310, opened 2023-11-03
Colored Cursors and Colored Selections

### LINKED ISSUE #202: Presence: colored cursors
As a user of VZCode in a live collaboration session with multiple people, I want to be able to tell whose cursor is whose based on the color of the cursor.

Google Docs does this.

VizHub2 does this - see https://github.com/vizhub-core/vizhub/blob/01cadfb78a2611df32f981b1fd8136b70447de9e/vizhub-v2/packages/neoFrontend/src/pages/VizPage/Body/Editor/CodeEditor/CodeArea/CodeAreaCodeMirror5/PresenceWidget/index.js#L4 - track down definition of `userColor` - there is an algorithm in here

aruan20:
Goal: to implement colored cursors and selections

Closes #202


aruan20:
First pass at making colored cursors and colored selections
- successfully imported and implemented the color hash
- implemented the colored cursors, but should find a better way to do it 
  - problem is that the styling for the cursor is in a function that doesn't have access to the user's color
- implemented the colored selections, but need to implement the mix-blend-mode for overlapping selections

curran (approved):
LGTM! I tried it locally and it works great. Thank you!

PR #314, opened 2023-11-03
Debounce Prettier on cursor and keyboard events

### LINKED ISSUE #157: Debounce auto-save with cursor movement
Currently, only changes trigger the debounce of auto-save.

As a user who is moving my cursor around actively, I don't want the auto-save to happen quite yet because I'm about to type something.

EvWhymark:
Closes #157 

Added the functionality to stop prettier from running on cursor move and keydown events. Still does not properly handle when others are changing text and waiting on the AI generation. I generally added a keydown event so if a user wants to have prettier run they will need to leave the editor idle (no keyboard inputs or mouse inputs). We can easily change it so it only does it on specific arrow keys if we want.

### DIFF
--- a/src/client/usePrettier/index.ts
+++ b/src/client/usePrettier/index.ts
@@ -128,6 +128,10 @@ export const usePrettier = (
       );
     }
 
+    //Stops prettier from running when you perform a mouse move or a keydown
+    window.addEventListener('mousemove', debouncePrettier);
+    window.addEventListener('keydown', debouncePrettier);
+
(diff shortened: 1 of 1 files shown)


curran (approved):
Excellent! Debouncing on all keyboard events is a great approach.

Interesting to debounce on all mouse move events... That hadn't occurred to me! It may be just right, but it might be too broad. Not sure. We can always remove that part later if it becomes an issue. Thank you for this!

Issue #31, opened 2023-02-07
syntax highlighting for remaining languages


curran:
* [X] .js
 * [X] .html
 * [X] .csv (do no highlighting)
 * [x] .css
 * [x] .md
 * [x] .json - just use JS highlighting
 * [x] .jsx - use JS highlighting

curran:
https://www.npmjs.com/package/@codemirror/lang-javascript#user-content-jsxlanguage

Might need to specify some additional options for JSX support.

PR #315, opened 2023-11-03
File renaming modal

DomaDude123:
Prompt the user with a modal when they attempt to create a new file.  When used with the createfile shortcut (alt-n) it creates a new file named "UnnamedFile".  In future versions users will be prompted with the same Modal when creating a file with the shortcut.

### DIFF
--- a/src/client/TabList/index.tsx
+++ b/src/client/TabList/index.tsx
@@ -20,7 +20,7 @@ export const TabList = ({
   setActiveFileId: (fileId: FileId) => void;
   openTab: (tabState: TabState) => void;
   closeTabs: (fileIds: FileId[]) => void;
-  createFile: () => void;
+  createFile: (fileName: string) => void;
 }) => {
   // Close the active tab on alt+w
   const handleKeyPress = useCallback(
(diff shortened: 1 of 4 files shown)


src/client/VZSidebar/CreateFileModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+            Enter the name for your file
+          </Form.Text>
+        </Form.Group>
+      </Modal.Body>
+      <Modal.Footer>
+        <Button
+          variant="primary"
+          onClick={handleRenameClick}
+        >
+          Rename
```
curran, line 76:
```suggestion
          Create File
```

curran (approved):
LGTM! I'll do a bit of cleanup then merge

src/client/TabList/index.tsx
```diff
@@ -30,7 +30,7 @@ export const TabList = ({
           closeTabs([activeFileId]);
         }
         if (event.key == 'n') {
-          createFile();
+          createFile('UnnamedFile');
```
curran, line 33: Interesting.... This will actually write the file to the file system, an will set up the editor with no file extension.

Good for a first pass! Related to #31 (that will need to happen for syntax highlighting to be correct after a the file gets renamed with a new extension)

curran (approved):

PR #316, opened 2023-11-06
Zap icon using shareDB to trigger AI Assist

Darmin-Geek:


src/client/AIAssistWidget/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+}: {
+  activeFileId: FileId;
+  shareDBDoc: ShareDBDoc<VZCodeContent>;
+  editorCache: EditorCache;
+}) => {
+  const handleClick = useCallback(() => {
+    // TODO make this happen
+    // See src/client/AIAssist.ts
+
+    console.log('AI Assist clicked');
```
curran, line 38:
```suggestion
```

src/client/AIAssistWidget/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  shareDBDoc,
+  editorCache,
+}: {
+  activeFileId: FileId;
+  shareDBDoc: ShareDBDoc<VZCodeContent>;
+  editorCache: EditorCache;
+}) => {
+  const handleClick = useCallback(() => {
+    // TODO make this happen
+    // See src/client/AIAssist.ts
```
curran, line 36:
```suggestion
```

src/client/AIAssistWidget/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  activeFileId,
+  shareDBDoc,
+  editorCache,
+}: {
+  activeFileId: FileId;
+  shareDBDoc: ShareDBDoc<VZCodeContent>;
+  editorCache: EditorCache;
+}) => {
+  const handleClick = useCallback(() => {
+    // TODO make this happen
```
curran, line 35:
```suggestion
```

curran (approved):
LGTM! Excellent work!

From here I would suggest, let's continue to improve this feature:

 * Having two different keyboard shortcuts feels like a temporary solution. Let's make it just one, `control-m`, to start and stop. This should be possible to do by inspecting the state of `shareDBDoc.data` and conditionally start or stop it depending on the current content there.
  * The UI should ideally allow the user to stop generation as well. As a first version, let's make it so that the icon changes to the text "stop generation" while generation is happening, then if you click it during that time it stops generation.

Issue #306, opened 2023-11-03
Error Overlay should Clear on Escape Key


curran:
As a user editing a file with a syntax error, I want to be able to easily dismiss the error by hitting the ESC key.

PR #321, opened 2023-11-07
Error Overlay should Clear on Escape Key

nghongwatsang:
#306

### DIFF
--- a/src/client/CodeErrorOverlay/index.tsx
+++ b/src/client/CodeErrorOverlay/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, useEffect } from 'react';
+import { useState, useCallback, useEffect } from 'react';
 import './style.scss';
 
 const enableErrorDismiss = true;
--- a/src/client/CodeErrorOverlay/index.tsx
+++ b/src/client/CodeErrorOverlay/index.tsx
@@ -30,7 +30,7 @@ export const CodeErrorOverlay = ({
     if (e.key === 'Escape') {
       handleCloseClick();
     }
-  }
+  }, [handleCloseClick]);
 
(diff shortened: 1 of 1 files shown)


src/client/CodeErrorOverlay/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState, useCallback, useEffect } from 'react';
+import React, { useState, useCallback, useEffect } from 'react';
```
curran, line 1:
```suggestion
import { useState, useCallback, useEffect } from 'react';
```

The `React` import is not required.

src/client/CodeErrorOverlay/index.tsx
```diff
@@ -18,6 +17,21 @@ export const CodeErrorOverlay = ({
...
+  useEffect(() => {
+    document.addEventListener("keydown", detectKeyDown, true);
+
+    return () => {
+      // Remove the event listener when the component unmounts
+      document.removeEventListener("keydown", detectKeyDown, true);
+    };
+  }, []);
+
+  const detectKeyDown = (e) => {
```
curran, line 29:
```suggestion
  const detectKeyDown = useCallback((e) => {
```

src/client/CodeErrorOverlay/index.tsx
```diff
@@ -18,6 +17,21 @@ export const CodeErrorOverlay = ({
...
+    if (e.key === 'Escape') {
+      handleCloseClick();
+    }
+  }
```
curran, line 33:
```suggestion
  }, [handleCloseClick]);
```

curran (approved):
LGTM!

Opportunity to leverage `useCallback` - I'll add that and merge.

PR #322, opened 2023-11-07
Better file renaming v2

DomaDude123:
Users are now prompted with the new file creation modal when they use the create file keyboard shortcut. 

Fixed small bug where previous filename would be in the new name field of subsequent modals.


curran (approved):
LGTM!

PR #323, opened 2023-11-07
285 open tabs as context for ai

Darmin-Geek:


curran (changes requested):
In general it's preferable to isolate the changes for a specific feature, rather than stacking the commits for multiple. Currently the tabs context is tangled up with the other changes around the ability to stop the stream, meanwhile I've added commits on top of the first PR to clean up that implementation.

Would you be able to please re-do this PR, branched off the current `main`, with only the tabs-as-context changes isolated? It's mainly copy-paste, but doing that would make it a lot easier to evaluate that work on its own. Thank you so much.

Issue #68, opened 2023-03-21
Presence: Usernames


Anooj-Pai:

curran:
Agree this would be nice to have.

Here's an example that does it, maybe we can draw some ideas from here https://github.com/share/sharedb/tree/master/examples/rich-text-presence

curran:
![image](https://github.com/vizhub-core/vzcode/assets/68416/67451b02-807b-4113-8478-87a08b7d54fe)

VSCode liveshare UI example

curran:
https://github.com/yjs/y-codemirror/blob/main/src/y-codemirror.js#L180C2-L180C36

curran:
To clarify scope of this issue - this is about displaying usernames near the presence cursors.

![image](https://github.com/vizhub-core/vzcode/assets/68416/420b310f-961a-4975-92d4-cbeca7918791)

We can probably do something similar to this implementation, which does the exact thing we want:

https://github.com/yjs/y-codemirror/blob/main/src/y-codemirror.js#L180C2-L180C36

PR #334, opened 2023-11-10
WIP Presence Usernames

aruan20:
Goal: show the username of the remote user near their cursor

Closes: #68


curran:
Very nice! I can see it working 
image
 I notice it does not work with selections. That would be neat to get working as well. It could show above the beginning of the selection in that case.

curran:
I'm ready to merge this, it looks great as a first version! Let me know if you feel it's ready of if you'd like to work on it more.

I do like the background color idea, but the text contrast is poor. Maybe try making the text black instead of light?

Also the positioning seems a bit off. Ideally it would show above the line (maybe 15-20 px higher than it is), but currently it shows in the middle of the line.

Excellent work!

aruan20:
I think it's ready to merge as a first version. Here's how it looks currently:
Screenshot 2023-11-10 at 4 15 51 PM
About the text contrast, it's currently set to black, but when there is a dark background, the text becomes hard to read as well. This could be a future improvement.
Another issue is that with the username displayed, some of the text on the file could get covered. A potential solution is to make the username appear only when you hover over the cursor.

PR #341, opened 2023-11-10
Validate Filename changes

DomaDude123:
Added functionality to check if a filename is valid before allowing users to proceed with the create file button.

### DIFF
--- a/src/client/VZSidebar/CreateFileModal.tsx
+++ b/src/client/VZSidebar/CreateFileModal.tsx
@@ -30,7 +30,7 @@ export const CreateFileModal = ({
     setNewName('');
   }, [newName, onRename]);
 
-  //returns true if filename isvalid, false otherwise
+  // Returns true if file name is valid, false otherwise.
   const validateFileName = useCallback((fileName : string) => {
     let regex = /^[a-zA-Z0-9](?:[a-zA-Z0-9 ./+=_-]*[a-zA-Z0-9])?$/;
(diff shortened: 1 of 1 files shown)


src/client/VZSidebar/CreateFileModal.tsx
```diff
@@ -30,6 +30,13 @@ export const CreateFileModal = ({
     setNewName('');
   }, [newName, onRename]);
 
+  //returns true if filename isvalid, false otherwise
```
curran, line 33:
```suggestion
  // Returns true if file name is valid, false otherwise.
```

src/client/VZSidebar/CreateFileModal.tsx
```diff
@@ -30,6 +30,13 @@ export const CreateFileModal = ({
...
+  const validateFileName = useCallback((fileName : string) => {
+    let regex = /^[a-zA-Z0-9](?:[a-zA-Z0-9 ./+=_-]*[a-zA-Z0-9])?$/;
```
curran, line 35:
```suggestion
    const regex = /^[a-zA-Z0-9](?:[a-zA-Z0-9 ./+=_-]*[a-zA-Z0-9])?$/;
```

If this was from source, it's best to include the link to where it was found.

Also, it would be good to list out in a comment what is meant by "valid".

curran (approved):
LGTM!

PR #342, opened 2023-11-10
TypeScript linter bug fix

### LINKED ISSUE #336: TypeScript Linter: Update File with Each Request
https://github.com/vizhub-core/vzcode/pull/335/files/e7cf729e914f3fc77216f0f906b268692d2dea7b#diff-e2b50b056500704b20f1ea54e0b788eaed9c52d4c207d093efce243e9cce694cR197

In `src/client/useTypeScript/worker.ts` line 197:

It's possible that a linting request is made before a completion request, and also before the debounced ShareDB updating logic has a chance to set the correct file content.

image

In fact while testing I received this error, indicating that the TS FS did not have the latest. We may want to update it here as well, on each request, just like we do for the completions.

EvWhymark:
Closes #336 

Thanks @curran, your suggestion worked! This gets rid of all the bugs that I found with my testing in the typescript linter. The content that the typescript server was running off of in the linter was out of date and we just needed to update the content with the shareDBDoc. I have also turned the feature flag for the linter on so it will be running in the codebase.

### DIFF
--- a/src/client/CodeEditor/getOrCreateEditor.ts
+++ b/src/client/CodeEditor/getOrCreateEditor.ts
@@ -37,7 +37,7 @@ import { typeScriptLinter } from './typeScriptLinter';
 
 // Feature flag to enable TypeScript completions & TypeScript Linter.
 const enableTypeScriptCompletions = true;
-const enableTypeScriptLinter = false;
+const enableTypeScriptLinter = true;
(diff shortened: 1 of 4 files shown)


curran (approved):
Hooray! Looks fantastic. Thank you.

PR #350, opened 2023-11-13
Start towards AI Assist middleware

curran:
When trying to integrate the AI Assist feature into VizHub, I ran into an issue where the server can't be listening to all documents at once, and triggering it in the middleware like this seems like a viable option instead.


curran:
Cancel this - it doesn't scale horizontally to multiple servers in a cluster.

PR #376, opened 2023-12-01
Rename Directory

jlewenthal2004:
Closes Issue #103


curran (changes requested):
The code looks decent, but I see no way to actually rename the directory while using VZCode.

The first step of this is to set `enableRenameDirectory = true;` on line 13 of `src/client/VZSidebar/Item.tsx`, then test the UX flow for renaming directories.

We also need to account for the change in the server, which is responsible for updating the file system to match the renamed files.

Also, please branch off the latest `main` in this repo itself. I see your fork is 156 commits behind vizhub-core:main. Thanks!

curran:
Closing as stale.

PR #393, opened 2023-12-05
Fix Autocomplete Cursor Bug

### LINKED ISSUE #362: Completions sometimes leave cursor on the left
After accepting a completion, the cursor stays on the left. It should go on the right.

Note: It does not happen all the time.

image

To reproduce:

```js
import { select } from "d3";
import { viz } fr
```

accept the completion "from".

EvWhymark:
Closes #362 

Autocomplete now forcefully moves the cursor to the correct position after doing a completion.


curran (approved):
LGTM!

PR #414, opened 2023-12-17
#110 respect .gitignore

### LINKED ISSUE #110: Ignore node_modules
Not sure what exactly this error means, maybe it's a symlink or something? But in any case, we should probably ignore `node_modules` for now.

```
> 19vitehmrtest@0.0.0 edit
> vzcode

node:fs:1437
  handleErrorFromBinding(ctx);
  ^

Error: ENOTDIR: not a directory, scandir '/home/curran/repos/stealth-prototypes/23BookScatterPlots/node_modules/is-wsl/node_modules/.bin/is-docker'
    at Object.readdirSync (node:fs:1437:3)
    at computeInitialDocument (file:///home/curran/repos/stealth-prototypes/23BookScatterPlots/node_modules/vzcode/src/server/computeInitialDocument.js:31:8)
    at file:///home/curran/repos/stealth-prototypes/23BookScatterPlots/node_modules/vzcode/src/server/index.js:24:25
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25) {
  errno: -20,
  syscall: 'scandir',
  code: 'ENOTDIR',
  path: '/home/curran/repos/stealth-prototypes/23BookScatterPlots/node_modules/is-wsl/node_modules/.bin/is-docker'
}

Node.js v19.4.0
```

natpbs:
don't include ignored file in the file tree, according to the contents of .gitignore and .ignore files

inspired by [ripgrep](https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#automatic-filtering) I set up vzcode to skip th files that match the patterns in the .gitignore or .ignore files it encounters when building the file tree. 

Differently from git or ripgrep, I'm not searching parent directoriees for ignore files and only take into account ignore files in the current working directory or it's descendants.

The implementation of the ignore pattern matching is based on [globby](https://github.com/sindresorhus/globby/blob/bd76374396886157357883b2223a342ff983b032/ignore.js#L28), with a few improvements. 

- when a (git)ignore file is found in a directory whose relative path includes special characters (for example the `**` directory found in `test/sampleDirectories/listFilesIgnoreEscape/`) the path is properly escaped when converting the path to be relative to the root of the file tree
- if a glob found in a (git)ignore file does not contain leading or middle slashes (i.e., any slash that is not the glob's last character), according to [git's docs](https://git-scm.com/docs/gitignore#_pattern_format) the pattern should match the name of files in any directory descending from the one where the corresponding gitignore file is found.

closes #110

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@codemirror/lang-markdown": "^6.2.3",
         "@codemirror/state": "^6.3.3",
         "@codemirror/theme-one-dark": "^6.1.2",
-        "@codemirror/view": "^6.22.1",
+        "@codemirror/view": "^6.22.2",
         "@lezer/highlight": "^1.2.0",
         "@replit/codemirror-indentation-markers": "^6.5.0",
         "@replit/codemirror-interact": "^6.3.0",
@@ -42,6 +42,7 @@
         "diff-match-patch": "^1.0.5",
         "dotenv": "^16.3.1",
         "express": "^4.18.2",
+        "ignore": "^5.3.0",
         "json0-ot-diff": "^1.1.2",
         "ngrok": "^4.3.3",
         "open": "^9.1.0",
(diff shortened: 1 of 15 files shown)


curran (approved):
Looks fantastic! Thank you so much for this work.

PR #415, opened 2023-12-18
Add shortcuts for manual prettier

curran:


src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -40,6 +40,7 @@ import { typeScriptLinter } from './typeScriptLinter';
 import { TabState } from '../vzReducer';
 import { keymap } from '@codemirror/view';
 import { basicSetup } from './basicSetup';
+import { isArrayBindingElement } from 'typescript';
```
curran, line 43: Accidental
```suggestion
```

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -294,7 +295,17 @@ export const getOrCreateEditor = ({
 
   // VSCode keybindings
   // See https://github.com/replit/codemirror-vscode-keymap#usage
-  extensions.push(keymap.of(vscodeKeymap));
+  // extensions.push(keymap.of(vscodeKeymap));
+  extensions.push(
+    keymap.of(
+      vscodeKeymap.map((binding) => {
+        if (binding.key === 'Enter') {
```
curran, line 302:
```suggestion
        // Here we override the Shift+Enter behavior specifically,
        // as that can be used to trigger a manual save/Prettier,
        // and the default behavior from the keymap interferes.
        if (binding.key === 'Enter') {
```

PR #417, opened 2023-12-20
make ignore file patterns configurable with environment variables

natpbs:
This is a follow-up to #414.
The associated issue #110 has already been closed, let me know if it's okay to reopen that one or a new one should be created. 

I fixed some corner cases and made the implementation a little faster, plus I added the option to configure which files are used as ignore files (by default files whose name is .gitignore, .ignore or .vzignore) and what ignore patterns are included by default.

I think these are useful improvements to the previous changes, but I'm open fors suggestions if you would prefer to revert some of these additions and only keep some. 

### Fix directory patterns
Directory patterns with a trailing slash (e.g., `.git/` or `node_modules/`) would correctly ignore the directory's contents, but not the directory itself. 

This would result in an empty directory in the file tree. I changed it so that the directory itself will be ignored and omitted from the file tree. 

### Performance improvement
In #414 I kept a list of all the patterns (as `string[]`) I encountered from parent folders and concatenated them all to create a new ignore object when a child directory included a new ignore file.

With this change, I reuse the parent's ignore object, which already includes all the patterns, and only add the new ones. 

This prevents the ignore library from having to parse the same pattern more than once, and also results in slightly simpler code. 

### Configuration with environment variables
I extracted the call to dotenv.config() to a separate module arc/server/setupEnv.js as per the following references.
https://www.npmjs.com/package/dotenv#how-do-i-use-dotenv-with-import
https://github.com/motdotla/dotenv/issues/133#issuecomment-255298822

This allows us to use environment variables defined in the `.env` file in other modules imported by SRC/server/index.js, such as featureFlags.js or the new config.js.

The file config.js for now just reads the values of the two environment variables BASE_IGNORE and IGNORE_FILE_PATTERN, setting default values if they are not present. 

I also updated featureFlags.js to read DEBUG_* environment variables instead of hard coded booleans, I find this more convenient when debugging a feature since it allows to set the feature flag in the .env file, which is ignored by git.

### DIFF
--- a/src/server/config.js
+++ b/src/server/config.js
@@ -1,0 +1,24 @@
+/**
+ *patterns describing which files should e comsidered as ignore files when creating the file tree.
+ *
+ * Format: one pattern per line, in gitignore format as defimed by https://git-scm.com/docs/gitignore#_pattern_format
+ */
+export const ignoreFilePattern =
+  process.env.IGNORE_FILE_PATTERN ??
+  `
+.vzignore
+.ignore
+.gitignore
+`;
+
+/**
+ * the base ignore patterns, that should be applied before any ignore file found when creating the file tree.
+ *
+ * These patterns apply even in the absence of an ignore file.
+ */
+export const baseIgnore =
+  process.env.BASE_IGNORE ??
+  `
+.git/
+node_modules/
+`;
--- a/src/server/computeInitialDocument.js
+++ b/src/server/computeInitialDocument.js
@@ -4,8 +4,10 @@ import { randomId } from '../randomId.js';
 import {
   enableDirectories,
   debugDirectories,
+  debugIgnore,
 } from './featureFlags.js';
 import createIgnore from 'ignore';
+import { ignoreFilePattern, baseIgnore } from './config.js';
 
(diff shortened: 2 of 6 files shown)


src/server/config.js
```diff
@@ -0,0 +1,24 @@
+/**
+ *patterns describing which files should e comsidered as ignore files when creating the file tree.
+ *
+ * Format: one pattern per line, in gitignore format as defimed by https://git-scm.com/docs/gitignore#_pattern_format
+ */
+export const ignoreFilePattern =
+  process.env.IGNORE_FILE_PATTERN ??
+  `
+.vzignore
```
curran, line 9: Love it!

src/server/config.js
```diff
@@ -0,0 +1,24 @@
...
+.ignore
+.gitignore
+`;
+
+/**
+ * the base ignore patterns, that should be applied before any ignore file found when creating the file tree.
+ *
+ * These patterns apply even in the absence of an ignore file.
+ */
+export const baseIgnore =
```
curran, line 19: Very nice! This feels right.

curran (approved):
LGTM! Thank you so much.

PR #422, opened 2023-12-26
#317 better development experience

### LINKED ISSUE #317: Better Developer Experience
As a VZCode developer, I want to have a smooth experience.

Particular pain points:
 * Currently when we start the server with `npm run test-interactive`, it also builds the frontend for production, which is not actually required for a dev environment that uses Vite's dev server
 * Currently, developers need to keep 2 processes running in order to develop, namely `npm run test-interactive` which runs the API server and also `npm run dev` which runs the Vite dev server.
 
Ideal solution:

 * Ideally there would just be a single command `npm run dev` that starts the API server, does _not_ build the frontend for production, and also starts the Vite dev server.

natpbs:
closes #317 

allow running the vite dev server and the vzcode backend with a single comand `npm run dev-fullstack`

### Implementation
The module `src/server/index.js` now exports a `createServer` function that can be used to start the vzcode backend. 

The code used to parse command line arguments and start the server when calling vzcode from the command line has been moved to `src/server/cli.js`, and the `bin` property in package.json is updated accordingly.

A custom plugin has been added to vite.config.js, which starts the vzcode backend in the current working directory when the vite dev server is started withe the mode `fullstack` (by using the --mode flag).

A new script `dev-fullstack` is present in package.json. The script starts the vite dev server with the --open flag and the full stack mode. As a consequence, a browser is opened showing the running (live-reloading) app on port 5173, while the vzcode backend runs in the background on port 3030. 

In this fullstack setup, connecting to port 3030 directly won't show the compiled application, since serving of static files is disabled in the vzcode backend unless it is started from cli.js.


curran:
Wow thank you so much for this great work!

I was doing some testing, almost ready to merge this, when I noticed that the app refreshes every time you modify a file _within VZCode_.

Perhaps we could add something that ignores changes in the `test` directory, so we don't have this behavior? Many thanks!

To reproduce:

```
npm run dev-fullstack
```

Edit a file, e.g. `index.html`.

Observe that the entire app refreshes - this is not desirable.

curran:
Closing as stale.

Issue #99, opened 2023-05-24
initialize the open folders 


curran:
If we eventually add a stateful URL that contains which file is open, initialize the open folders to have the folders that contain that file automatically open on page load

curran:
reproduction link: https://vizhub.com/curran/17cc1386ccf447edb25b68909262a823?edit=files&file=data_chapters%2FbasicViz.js

Current behavior:

![Image](https://github.com/user-attachments/assets/8aab91b6-6f90-406d-b176-703353c2a63b)

Desired behavior:

![Image](https://github.com/user-attachments/assets/2e5d2896-8a3e-4f97-999a-29580bca5013)

PR #423, opened 2023-12-27
#100 deep linking

### LINKED ISSUE #100: Deep Linking
Deep Linking for:

 * [ ] The ordered list of open `FileIDs` corresponding to open tabs
 * [ ] Highlighted line number e.g. `https://vizhub.com/nitanagdeote/187ce4be0e3a4c1bbe06dc3a5bee3c40?edit=files&file=index.js#L13`

Image

### LINKED ISSUE #168: Active file in URL
As a user of VZCode, I want to be able to share a link with someone that captures which file I currently have open, so that the receiver of the link does not need to find the file and open it themselves.

As a user of VZCode who receives a link from someone else, I want to be able to open the editor with a specific file open, so that I don't need to search for it after having opened the link.

natpbs:
closes #100
closes #168 
related to #99 

save open tabs to the URL.

For each open tab, a search parameter is added specifying

 - the file id (not the path)
 - whether it's a transient r permanent tab
 - whetherthe tab is currently active.

On page load, the tabList saved in the URL is restored. 

### TODO

- [ ] Validate the open tabs file IDs against the list of available files.
- [ ] Find a way to handle restarts of the backend, which cause file IDs to be recreated from scratch. 
  - Client side: save the open paths somewhere (URL, history state, storage) and use those instead of the file IDs when none matches.
  - Server side: create a `.vzstate` or `.vzcode/state.json` file, which lists the loaded file tree and the corresponding file IDs and keep the file updated after file moves / renames. 

    When the server restarts, it finds that file and refuses the same file IDs for the corresponding paths instead of generating fresh ones.
 -[ ] Save the cursor's line, at least for the active file.


curran:
Thanks for this work!

It's unfortunate that this change is stacked on the other change related to "use vite mode 'fullstack'".

There was an issue with that full stack mode, which is now preventing merge of this PR.

Ideally each PR would branch off `main`, so they don't end up containing commits from other PRs.

Is there any chance you could consider doing this again, but branching off `main`?

Also, I'm a fan of `react-router` and would like to leverage that for the deep linking feature.

Thanks again!

package.json
```diff
@@ -4,17 +4,18 @@
   "description": "Multiplayer code editor system",
   "main": "src/index.ts",
   "type": "module",
-  "bin": "src/server/index.js",
+  "bin": "src/server/cli.js",
```
curran, line 7: CLI changes in the deep linking PR is unfortunate.

src/client/VZCodeContext.tsx
```diff
@@ -172,6 +178,52 @@ export const VZCodeProvider = ({
     setUsername,
   } = useActions(dispatch);
 
+  // initialize open tabs from the url search parameters
+  useEffect(() => {
+    const search = new URLSearchParams(
```
curran, line 183: Very nice! I would prefer to use React Router, so that we can leverage their various utilities.

src/client/VZCodeContext.tsx
```diff
@@ -172,6 +178,52 @@ export const VZCodeProvider = ({
...
+    );
+    if (search.size === 0) {
+      return;
+    }
+    const tabList: TabState[] = [];
+    let activeFileId = null;
+    for (const [fileId, tabState] of Array.from(
+      search.entries(),
+    )) {
+      const isTransient = tabState.includes('t');
```
curran, line 194: Looks like a clever encoding scheme. Ideally we could have some unit tests for encoding and parsing of these.

curran (changes requested):

natpbs:
Thank you for the feedback, I'll undraft the pull request after I incorporate all the suggestions.

package.json
natpbs: sorry for mixing up my branches, I rebased on main and soon will change also #422 to mitigate the change.

curran: Thank you so much!

curran:
Merged. Thank you so much @natpbs !

src/client/VZCodeContext.tsx
```diff
@@ -172,6 +187,29 @@ export const VZCodeProvider = ({
     setUsername,
   } = useActions(dispatch);
 
+  // save the opem tabs to the url search parameters
+  const tabListTimeoutId = useRef<number>(null);
+  useEffect(() => {
+    // debounce: only update the url when the tabList stops changing
```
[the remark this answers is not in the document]
curran: Aha! I deleted this whole chunk of code as it was replaced with `usePersistTabs`. See https://github.com/vizhub-core/vzcode/pull/436/files

curran:
Released! https://github.com/vizhub-core/vzcode/releases/tag/v0.50.0

Issue #430, opened 2023-12-31
Number Dragger should work outside editor


curran:
As a user editing a number with the interactive number dragger widget, I want to be able to continue dragging outside the code editor, so that I can get a greater range.

curran:
Related to https://github.com/replit/codemirror-interact/issues/16

curran:
This code works, but has a bug when the number of digits change:

```js

// a rule for a number dragger
{
  // the regexp matching the value
  regexp: /(?<!\#)-?\b\d+\.?\d*\b/g,
  // set cursor to "ew-resize" on hover
  cursor: 'ew-resize',
  // change number value based on mouse X movement on drag
  // This works fine with the 999 to 1000 transition
  // onDrag: (text, setText, e) => {
  //   if (onInteract) onInteract();
  //   const newVal = Number(text) + e.movementX;
  //   if (isNaN(newVal)) return;
  //   setText(newVal.toString());
  // },

  // This DOES NOT works fine with the 999 to 1000 transition
  // TODO figure out why
  onClick(text, setText, e) {
    let newVal = Number(text);

    const handleDrag = (e: MouseEvent) => {
      if (onInteract) onInteract();
      newVal += e.movementX;
      if (isNaN(newVal)) return;
      setText(newVal.toString());
    };
    // add an event listener to the document for mouse move
    document.addEventListener(
      'mousemove',
      handleDrag,
    );

    // when the mouse is released, remove the event listener
    document.addEventListener('mouseup', () => {
      document.removeEventListener(
        'mousemove',
        handleDrag,
      );
    });
  },
},
```

If it starts with, say 900, then is dragged past 1000, each new change adds a digit. Likewise if it decreases from say 100 to 99, subsequent characters get deleted. Any ideas why that might be happening?

I think it has to do with this logic:

```js
  updateText(target) {
    return (text) => {
      view.dispatch({
        effects: setInteract.of({ ...target, text }),
        changes: {
          from: target.pos,
          to: target.pos + target.text.length, // The target length has changed, but how can we update it?
          insert: text,
        },
      });
    };
  },
```

curran:
I realized that this is almost the exact same issue as #426 at its core. I bet we could solve both at once.

curran:
This feature requires that we transform the insertion cursor positions (start and end) by all ShareDB ops.

curran:
Here's what I tried, from 

```js
  // This DOES NOT works fine with the 999 to 1000 transition
[...]

PR #431, opened 2023-12-31
Attempt number dragger off editor

curran:
An attempt towards #430


curran:
Closing as stale.

This feature requires that we transform the insertion cursor positions (start and end) by all ShareDB ops.

PR #475, opened 2024-01-30
457 resizer blocks scrolling

### LINKED ISSUE #457: Resizer Blocks Scrolling - Can't grab the vertical scrollbar in editor
image

The resizer on the right appears to be offset slightly to the left, which is a bug that is preventing the user from scrolling by grabbing the scroll bar.

Ideal solution: Move the resize grabber to the right so that the user can scroll by grabbing the scroll bar handle, and ALSO resize if hovering slightly to the right of the scroll bar.

viratkohli2011:
Added isDragging condition that adjusts the position when dragging to keep it aligned

Closes #457


curran:
Your changes are not quite right. There is an offset on the thumb (blue portion) that should not be there. We need to move over the outer element (class `vz-resizer` - yellow portion here) so that it does not overlap with the scrollbar handle:

image

Also the blue part should remain centered on the dividing line between the two panes.

Two subtasks:

 * Introduce an offset on the parent element (class `vz-resizer`)
 * Introduce an opposing offset on the child element (class `vz-resizer-thumb`)

curran:
There are merge conflicts now. I would suggest to start a new PR from `main` to make things easier, since the first change is so small and we don't want to keep it anyway.

Good luck!

PR #484, opened 2024-02-03
Stable layout

CJary:
I added a maximum height for the class of file-or-directory.

### DIFF
--- a/src/client/VZSidebar/styles.scss
+++ b/src/client/VZSidebar/styles.scss
@@ -1,7 +1,7 @@
 .vz-sidebar {
   height: 100%;
   color: #d6d9de;
-  background: #33415b;
+  background: #1f2022;
   display: flex;
   flex-direction: column;
   justify-content: space-between;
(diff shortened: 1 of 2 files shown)


src/client/VZSidebar/styles.scss
```diff
@@ -22,6 +22,7 @@
   }
 
   .file-or-directory {
+    height: max(30px);
```
curran, line 25: In this context, max(30px) effectively just sets the height to 30px, because there is no other value to compare it against. Therefore, it can be simplified by directly setting the height to 30px without using the max() function.

```suggestion
    height: 30px;
```

curran (approved):
LGTM. Thanks!

curran:
Thanks for this! However, as I worked through some other styling issues, I realized that setting the height like this not ideal, because then the height does not flow naturally from the content, which I would prefer it did so that we can get things like padding more consistent.

If you don't mind, can you please try working towards a solution that modifies the style of the input field?

This state is fine and can remain the same:

image

This is the state we need to work on:

image

It shouold be possible to fix it with only CSS changes that apply to the input field. You may want to disable the rename functionality for testing, so that you can get the input field to appear and not go away as you tweak the CSS. Good luck!

PR #507, opened 2024-02-07
Ai assist spinner

CJary:
Worked with demhos && NghiaMinhHuynh!

Added the implementation of the Spinner in a new file called Spinner.jsx.

Added the Spinner's CSS into the styles.scss.

Changed the return of the index.tsx react component. Made it so that it will render either the AI assist icon or the spinning. If the API id is not NULL it will show the spinner, else it will show the ai assist icon.


curran (approved):
Looks great! Works great! Thank you so much.

image

Issue #505, opened 2024-02-05
Rename and Trash Icons stay after hover


chinm4:
When navigating up on the sidebar over the Rename and Trash icons the icons are staying there permanently until hovered again
![afterhoveringuponeachfilesicons](https://github.com/vizhub-core/vzcode/assets/90073238/42df2d7f-f862-4c4f-b0e3-fb5fb0e0f3f5)

curran:
Are you on the latest `main`?

curran:
Interesting... I cannot reproduce it.

curran:
If we change the hover display for those buttons to be done with CSS rather than JS, I think that would solve this issue.

Otherwise, we may need to listen for more types of mouse events. Maybe it's not getting the mouse out, or there may be some issue with event bubbling out of the tooltip.

curran:
Possibly related to these new changes

https://github.com/vizhub-core/vzcode/pull/493

curran:
Another idea on this: the problem seems to be caused by mouse events on the tooltip. What if we just disable all pointer events for the tooltip? That could be an easy fix.

curran:
Here's an idea for a fix: move the state of the hovered item up one level in the component tree.

Or maybe we could just use CSS to show/hide the icons when you hover on the parent.

curran:
Finally fixed in https://github.com/vizhub-core/vzcode/pull/617 !

PR #522, opened 2024-02-09
issue505 first draft

MandingoBrandon:
https://github.com/vizhub-core/vzcode/pull/new/issue505

### DIFF
--- a/src/client/VZSidebar/Item.tsx
+++ b/src/client/VZSidebar/Item.tsx
@@ -1,21 +1,10 @@
-import {
-  useState,
-  useCallback,
-  useRef,
-  useEffect,
-} from 'react';
-import { EditSVG, FileSVG, TrashSVG } from '../Icons';
-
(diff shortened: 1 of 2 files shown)


curran (changes requested):
This PR has lots of additional changes that are unrelated to the issue at hand. The existing comments need not be deleted. 

I don't understand why the new state fields were added for tracking when the buttons are hovered.

Here's what I would suggest:

 * Start fresh by branching off `main`
 * Try making a small surgical change that removes the `const [isHovered, setIsHovered] = useState(false);` line
 * Try making the icons show and hide by leveraging CSS, something kind of rule like "show the icons when the parent is hovered.

Thanks!

curran:
Another idea: there's only ever _one_ hovered item. What if we track that state higher up in the component tree? That might solve it.

curran:
#505

PR #538, opened 2024-02-13
Intelligent Autocompletions: Styling v1

CJary:
Worked with hosand

I applied the CSS but removed the padding because it creates a random box. However, I'm confused about what you mean by the palette's color. So I tried to match the color with the VZSidebar.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "@replit/codemirror-interact": "^6.3.0",
     "@replit/codemirror-lang-svelte": "^6.0.0",
     "@replit/codemirror-vscode-keymap": "^6.0.2",
-    "@rollup/rollup-darwin-arm64": "^4.10.0",
     "@teamwork/websocket-json-stream": "^2.0.0",
     "@typescript/vfs": "^1.5.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -77,6 +77,7 @@
     "@replit/codemirror-interact": "^6.3.0",
     "@replit/codemirror-lang-svelte": "^6.0.0",
     "@replit/codemirror-vscode-keymap": "^6.0.2",
+    "@rollup/rollup-darwin-arm64": "^4.10.0",
```
curran, line 80:
```suggestion
```

This will break `npm install` for me and others not on that platform.

src/client/CodeEditor/style.scss
```diff
@@ -24,4 +24,21 @@
   .cm-tooltip-lint {
     max-width: 500px;
   }
+
+    /* Style the completion tooltip container */
+  .cm-tooltip.cm-tooltip-autocomplete {
```
curran, line 29: Oh wow nice! This is way simpler than I thought it would be. This is fantastic!

curran (approved):
Excellent!

Issue #477, opened 2024-01-31
TypeScript Linting: Ignore "any type" warnings


curran:
This is such crap I don't want to see it

![image](https://github.com/vizhub-core/vzcode/assets/68416/9b9cc5f0-704d-496e-bd2e-a909858904ab)

Repro:

```
let mic;
```

PR #543, opened 2024-02-16
anytype warning

CJary:
Added, "noImplicitAny": false to tsconfig.json

### DIFF
--- a/src/client/useTypeScript/worker.ts
+++ b/src/client/useTypeScript/worker.ts
@@ -234,6 +234,10 @@ onmessage = async ({ data }) => {
         // "Cannot find module 'd3' or its corresponding type declarations."
         const LINT_ERROR_CODE_IMPORT = 2307;
 
+        // This code is for errors like:
+        // "Variable 'mic' implicitly has type 'any' in some locations where its type cannot be determined."
+        const LINT_ERROR_CODE_ANY_TYPE = 7034;
+
         if (debug) {
           console.log(tsErrors);
         }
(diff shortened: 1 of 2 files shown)


curran:
What issue does this close?

CJary:
If I implemented it correctly, I hope it will fix TypeScript Linting: Ignore "any type" warnings
[#477](https://github.com/vizhub-core/vzcode/issues/477)

curran:
Ah, I see! The change you have made here relates to the config of the VZCode source code itself.

That issue is about this kind of error appearing inside the actual VZCode editor itself.

So to reproduce this, you would run `npm run dev`, then open up a file, then paste the reproduction code, then see the error. To remove that error, we'd need to make a source code change to filter out some of the linter warnings based on the error code. I believe this is already being done, and we probably can solve it by just adding another code to the list that we check already.

See https://github.com/vizhub-core/vzcode/blob/main/src/client/useTypeScript/worker.ts#L240

CJary:
Commit a new version 2/16 4:06pm. Restore the tsconfig.json file. Added the error code into the filter for the array tsErrors. 

Error code was referenced on this website I found:
https://github.com/microsoft/TypeScript/blob/v5.4-beta/src/compiler/diagnosticMessages.json

curran:
Amazing! Thank you so much. I will test it out.

curran:
I'm so happy this is getting resolved, it came up just today when I was recording a lecture video:

image

curran:
Tested it out, and it works! Thank you so much. This is a great improvement to the editor experience.

image

PR #554, opened 2024-02-25
Iss427

MandingoBrandon:
https://github.com/vizhub-core/vzcode/pull/new/iss427

I have the click on color picker working, but with the dragging in the picker and it changing the hex value is giving me more of an issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "@replit/codemirror-interact": "^6.3.0",
         "@replit/codemirror-lang-svelte": "^6.0.0",
         "@replit/codemirror-vscode-keymap": "^6.0.2",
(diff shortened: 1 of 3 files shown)


curran:
Thank you @MandingoBrandon for this work!

It's an interesting idea to have the mouse movement itself control the colors, but that's not really what we're going for here. Apologies if the original issues was not clear, but the goal with this issue is to make the color picker appear when you click on the colored circle. In other words, the existing interaction for opening the color picker (that happens when you use Alt+click on the hex value) should also happen when you click on the circle (without holding down Alt). Does that make sense?

This PR has a great start in that it does detect when you click on the circle. From there, we want to tie into the other logic around opening up the color picker and getting the values out.

curran:
It's a great attempt, but not quite feasible I think.

Issue #410, opened 2023-12-15
Connection Status


curran:
As an editor using VZCode in a Web application, I want to know when the connection has been lost.

![image](https://github.com/vizhub-core/vzcode/assets/68416/3eee09ad-5d20-4b4c-b3a6-a53de5dae33e)

![image](https://github.com/vizhub-core/vzcode/assets/68416/31b86a1e-8aba-4c27-b3a8-1ce75c9822e9)

PR #555, opened 2024-02-26
Closes #514

### LINKED ISSUE #514: Configurable Font
Enhancing the editor with a feature that allows users to customize the font provides a personalized and comfortable coding experience. This improvement would cater to individual preferences and accessibility needs, enabling users to select the font style and size best suits their visual requirements for optimal readability and productivity.

Akshithpottigari:
Using CSS variables for font family and the font size was easy to handle globally, and they were already present in the root, so no need to declare new variables and handle them differently, and also became handy for us to handle them throughout the app. We can add more fonts and more font sizes simply by loading them in the fonts.ts and font sizes array and we are good to go.

I have implemented #410 too, for detecting the connection status too. 

We can close #514 and #410

Thanks in advance : )

### DIFF
--- a/src/client/VZSettings.tsx
+++ b/src/client/VZSettings.tsx
@@ -2,6 +2,7 @@ import { useCallback, useContext, useRef } from 'react';
 import { Button, Modal, Form } from './bootstrap';
 import { themes } from './themes';
 import { VZCodeContext } from './VZCodeContext';
+import { fonts } from './Fonts/fonts';
 
 export const VZSettings = ({
   enableUsernameField = true,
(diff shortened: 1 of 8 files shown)


curran (approved):
Very nice! Thank you so much!

In future it would be best to address different issues in different PRs, but these changes are all good.

Some things to consider for future iterations on these features:

 * Ideally if a user selects a font and size, that should persist using localStorage, so that it's there when they refresh the page
 * The connection status is good, but does not handle the case of switching from "Saving..." to "Saved", which is something I'd really like to see in a future iteration

Thanks again!

src/client/VZSettings.tsx
```diff
@@ -77,6 +93,44 @@ export const VZSettings = ({
...
+
+        <Form.Group className="mb-3" controlId="formFork">
+          <Form.Label>Font</Form.Label>
+
+          <select
+            className="form-select"
+            onChange={handleFontChange}
+          >
+            {fonts.map((font) => (
+              <option key={font} value={font}>
```
curran, line 105: This does not show the previously selected option when you close the modal then re-open it.

curran: Indeed, it is correct! Now that the first version is ready, the next steps become clear.

I created a new issue to track the persistence: https://github.com/vizhub-core/vzcode/issues/562

curran: Oh but this comment was related to just the modal closing and re-opening, within a single session. I solved it in https://github.com/vizhub-core/vzcode/pull/559/files#diff-3cc8b6f15ff7e81275b96ef7bc6d64e13b89785525e651ccc6800b5da75aefa9R156

PR #575, opened 2024-03-09
drag and drop v1

CJary:
Implemented a new function called createFileWithContent in the useFileCRUD.ts and added it into the VZCodeContext.tsx for the DragAndDrop.jsx.

DragAndDrop.jsx is the component for a file being dragged over and created along with the content inside.

The frontend menu for the DragAndDrop is done by the VZSideBar's styles.scss.


curran (approved):
Excellent! Thank you so much! This works perfectly. I added a commit to clean up a few minor things.

As a follow-up to this one, I think it would make sense to treat the entire sidebar as the drop zone, rather than have a specific drop zone called out in the UI. I think that would be cleaner in terms of UI, and also that's what other editors like VSCode do. I've created a new issue for that here: https://github.com/vizhub-core/vzcode/issues/577

PR #576, opened 2024-03-09
Better-Styling-for-Rename

NghiaMinhHuynh:
I update the styles.scss file in the rename-input classname for better styling. I also add the FileSVG logo when we click to rename it keep the icon on the left while renaming.


curran:
Thank you so much for this work! It looks great, and I've merged it into `main`. It's part of the current release.

PR #587, opened 2024-03-15
drag and drop with entire sidebar

CJary:
I made the DragAndDrop.jsx into a custom hook that returns just the necessary functions and components. 

I added the custom hook to the index. jsx for the VZSidebar.

I edited the CSS slightly so that it's a solid border instead of dashed. In my opinion, it looks better.


curran (approved):
Beautiful! Thank you so much for working on this.

PR #607, opened 2024-03-22
Fixes Issue #600, Bug with Directory Deletion

### LINKED ISSUE #600: Directory Deleting Bug: Deleting a Directory deletes ALL Directories that start with the same name
If you try to delete a directory with a name such as "directory", it also deletes all directories at the same level with names "directory____" where "____" is any characters.
Example, Deleting "Directory" also deletes "Directory1" and "Directory2"...

enuessle:
Small change to make sure it only deletes the exact directory named
Closes #600


curran (approved):
Very nice! Thank you.

PR #633, opened 2024-03-29
Adds Validation to File Creation Keyboard Shortcut

### LINKED ISSUE #632: Bug: Pressing 'Enter' in the Create File Modal gets around the file name validation check
In the File Creation Modal, the 'Create File' button is disabled when the file name is not valid. The keyboard shortcut to create a file with 'Enter' is not disabled. This let's you create files with invalid names, often breaking the sync.

Example: Insert an emoji 😀 as a file name, and pressing 'enter' bypasses the validation check and creates a new file with 😀 as it's name, which doesn't work in the file system, causing issues and disconnecting the editor.

enuessle:
Closes #632

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,10 @@
     {
       "name": "Evan Whymark",
       "url": "https://github.com/EvWhymark"
+    },
+    {
+      "name": "Ethan Nuessle",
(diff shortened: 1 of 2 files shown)


curran (approved):
Very nice! Thank you.

PR #643, opened 2024-04-02
F2-Rename Draft

demhos:
This is what I have so far, not sure what else to add to context to make it enter the input box

### DIFF
--- a/src/client/VZCodeContext.tsx
+++ b/src/client/VZCodeContext.tsx
@@ -3,6 +3,8 @@ import {
   useCallback,
   useReducer,
   useState,
+  useEffect,
+  useRef,
 } from 'react';
 import {
   FileId,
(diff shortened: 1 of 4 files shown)


curran (changes requested):
The overall approach of moving the rename state to the context is good. However, it's only partially done, because the same logic still exists in `src/client/VZSidebar/Item.tsx`.

To wrap this up, I would suggest to remove the rename state completely from `Item.tsx`, and make sure the original rename logic (from clicking the rename button) uses the state from the context.

Also, I'm doubtful of adding the field `name` to the reducer state. Please consider making it more clear what it is for. I get that it is required to keep track of which file is being renamed, but it may be better to use a `fileId` instead, so that it's robust to renames? Maybe it could be called something like `fileIdBeingRenamed` or something.

Thanks for the work on this, it's almost there! Good luck.

demhos:
I've removed the states from Item.tsx and changed the name of the name in reducer to fileIdBeingRenamed for better clarification.

curran:
Looks great! One thing I notice is that when I actually test it out, I press F2 when a file is selected and I get the following error:

image

I'm curious, when you select a file and press F2, does this error not happen for you?

I think the problem is that the text field is not set up yet in the DOM, so it tries to auto-focus it too soon.

PR #644, opened 2024-04-02
draganddrop for dir

CJary:
This is a draft pull request. 

I implemented the base of the directionary process. So far:
Case 1: Regular files being dropped still works. 
Case 2: Folders when dropped seems to cause a npm error in the server/index.js file. Visually it shows up in the VZ editor however it will lose connection.

### DIFF
--- a/src/client/VZCodeContext.tsx
+++ b/src/client/VZCodeContext.tsx
@@ -54,7 +54,7 @@ export type VZCodeContextValue = {
   docPresence: any;
 
   files: Files | null;
-  createFile: (fileName: string, text?: string,) => void;
+  createFile: (fileName: string, text?: string) => void;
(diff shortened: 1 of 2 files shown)


src/client/VZCodeContext.tsx
```diff
@@ -54,7 +54,7 @@ export type VZCodeContextValue = {
   docPresence: any;
 
   files: Files | null;
-  createFile: (fileName: string, text?: string) => void;
+  createFile: (fileName: string, text?: string,) => void;
```
curran, line 57:
```suggestion
  createFile: (fileName: string, text?: string) => void;
```

curran (approved):
Amazing! Thank you so much.

The server-side error may be solved in https://github.com/vizhub-core/vzcode/pull/647

PR #645, opened 2024-04-02
Persist Font

demhos:
I made it so the font and font size will remain after reloading the tab


curran (approved):
Very nice! Looks great! Thank you so much.

PR #654, opened 2024-04-09
Tell if run icon is pressed

demhos:
Made it so that the run icon spins when it is pressed


curran (approved):
LGTM!

This is cool! However, ideally it should also animate when the keyboard shortcuts are pressed. That can be done as a follow-on pull request perhaps.

PR #656, opened 2024-04-12
Ai assistclaude

CJary:
Review the API request on claude3: https://docs.anthropic.com/claude/reference/complete_post
Adjusted the startAIAssist.ts with the new body request that claude3 wants. Still missing the API key for Claude so couldn't test it. But this is what I have so far.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.88.0",
       "license": "MIT",
       "dependencies": {
+        "@anthropic-ai/sdk": "^0.20.2",
         "@codemirror/autocomplete": "^6.15.0",
         "@codemirror/lang-css": "^6.2.1",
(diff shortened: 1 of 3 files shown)


curran (changes requested):
Excellent start! A few key changes to the overall approach are needed:

 * Based on the way VZCode is set up, the streaming AI response is added server-side, so the client side of this should ideally remain untouched
 * We want to use the streaming variant of the API, so we don't need to wait for the entire response
 * I think it should be possible to get an API key on a free Anthropic account, but I will investigate a little more and, if needed, I can provide an API key for you to use while testing.

Thank you so much for this work!

curran:
Closing as we can now use OpenAI-compatible endpoints that serve Claude, such as OpenRouter.

PR #657, opened 2024-04-12
Duplicate Filename Prevention on File Creation

### LINKED ISSUE #653: Bug: You can create files with duplicate names, which then deletes the originanal file.
In the VZCode interface, naming file to a name that already exists works, and creates a new file in VZCode.
Two Index js

On the file system however it just ends up deleting all the contents of the original file, which goes against what the VZCode interface is saying
Deletes Contents

Ideally it should work like VSCode and not let you name something if it already exists, or just add some numeral at the end to mark it as separate and not break things.

enuessle:
closes #653


curran (approved):
LGTM!

Issue #189, opened 2023-09-22
Widgets: Position of Color Picker


curran:
Current behavior: the color picker appears in the upper left of the display.

Desired behavior: it should appear next to the color in the code similar to https://github.com/replit/Codemirror-CSS-color-picker

![image](https://github.com/vizhub-core/vzcode/assets/68416/1d7570e5-0443-4f52-8ae8-df1b338826ad)

PR #664, opened 2024-04-21
Fix Color Picker Position

brandonbalchand:
https://github.com/vizhub-core/vzcode/pull/new/issue189


curran:
Oh nice! This looks like related to https://github.com/vizhub-core/vzcode/issues/189

PR #668, opened 2024-04-24
first check for desktop app electron

brandonbalchand:
https://github.com/vizhub-core/vzcode/pull/new/DesktopApp

### DIFF
--- a/electron-main.mjs
+++ b/electron-main.mjs
@@ -1,0 +1,54 @@
+import { app, BrowserWindow } from "electron";
+import path from "path";
+import express from "express";
+import cors from "cors";
+import { fileURLToPath } from 'url'; // Import fileURLToPath function
+
+const __filename = fileURLToPath(import.meta.url); // Get the current module's filename
+const __dirname = path.dirname(__filename); // Get the directory name
+
+const localServerApp = express();
+const PORT = 5173;
+
+const startLocalServer = (done) => {
+  localServerApp.use(express.json({ limit: "100mb" }));
+  localServerApp.use(cors());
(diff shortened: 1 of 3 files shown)


curran (changes requested):
This is amazing! Thank you so much for working on this.

I have a few points of feedback:

 * When I run `npm run dev`, I see the following error:
image
 * I would prefer not to change the existing scripts in `package.json`, but rather introduce a new one for Electron
 * Let's keep `"main": "src/index.ts",` for now, since I believe that impacts how we can import from the package. I'm not sure what the benefit of changing this is.
 * The usage of `cors` seems like it may not be strictly necessary. Perhaps it is, but something worth looking into, if we can remove it as a dependency.

curran:
To summarize the additional changes I'd like to see before merging this:

 * Please revert `npm run dev` to the same as it was before, so that I and others can continue to develop VZCode without invoking Electron
 * Please add a separate script that starts up the electron app and server, and ideally document in the README the exact commands recommended to start the electron app
 * Please revert the `"main": "src/index.ts"`
 * Please check if it can still work without the `cors` dependency

Thank you so much!

icoco:
[maybe pkg also is another option ](https://github.com/icoco/vzcode/tree/anvil) 
still not done

curran:
Whoah thank you @icoco !

I'd still like to make this happen. Not sure what the easiest path is, but it would be great to be able to do this:

```
npm install -g vzcode
cd my-local-source-code-dir
vzcode .
```

then have the app open as an Electron app to edit my files locally.

PR #703, opened 2024-06-04
uncommented automatic window opening code and added .history to .gitignore

michaelhelper:
The code which automatically opens a window had been uncommented and the functionality restored.
 Added .history to .gitignore


curran (approved):
LGTM! Thanks!

PR #704, opened 2024-06-04
Set color hash lightness to 0.75

michaelhelper:
Set color hash lightness to 0.75.

Take a look at it and let me know if the lightness value needs to be adjusted. In the limited testing I preformed the colors looked good to me.


curran (approved):
LGTM! Thank you.

PR #706, opened 2024-06-05
Preserve scroll across files

jeffrey-asm:
The proposed changes will ensure that the current scroll position for the active file will be cached when removing the editor from the DOM. The cached scroll position is applied when switching between files and opening closed files.

### DIFF
--- a/src/client/useEditorCache.ts
+++ b/src/client/useEditorCache.ts
@@ -5,7 +5,7 @@ import { EditorView } from 'codemirror';
 export type EditorCacheValue = {
   editor: EditorView;
   themeCompartment: any;
-  position?: number;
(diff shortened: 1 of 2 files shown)


src/client/useEditorCache.ts
```diff
@@ -5,6 +5,7 @@ import { EditorView } from 'codemirror';
 export type EditorCacheValue = {
   editor: EditorView;
   themeCompartment: any;
+  scrollPosition?: number;
```
curran, line 8: Brilliant idea attaching this to the editor cache value!

curran (approved):
Wow this works perfectly! Thank you so much!

PR #711, opened 2024-06-07
Fixed issue with typescript error for object reassignment (code 2739))

DaCatDude:
Added code 2739 to list, added .ts test file for specific error


curran (approved):
Perfect!!! Thank you so much.

PR #713, opened 2024-06-07
Update package.json

michaelhelper:
Added my name and url to package.json - (Michael Halpern)


curran (approved):
Excellent! We can add more folks in future PRs.

PR #714, opened 2024-06-07
655 ignore ts lint about unknown reopen

DaCatDude:
Turns out the specific error code for this one was 18046 not 2571 for type unknown. I reopened it in a different branch because the old one was something like 56 commits behind

Changed TSErrortest to reflect new test


curran (approved):
LGTM!

PR #717, opened 2024-06-07
Jump to definition

jeffrey-asm:
This implementation will allow users to use CTRL + Click when targetting an identifier within the current code editor to jump to its definition for various identifier types, such as variables, methods, and properties.


curran (approved):
LGTM!

PR #718, opened 2024-06-07
Follow user draft, users' scrolls follow the other's cursor around the same file. Works for clicking and arrow-key movement.

DaCatDude:
Please let me know if this works, I did some testing in a long javascript file, but I havent tested it anywhere else. As far as I know, it should work without many issues.

### DIFF
--- a/src/client/CodeEditor/json1PresenceDisplay.ts
+++ b/src/client/CodeEditor/json1PresenceDisplay.ts
@@ -109,7 +109,7 @@ export const json1PresenceDisplay = ({
             }
             if (view.state.doc.length >= from) { // Ensure position is valid
               this.cursorPosition[id] = from; // Store the cursor position, important to run if we cant get the regular scroll to work
-              console.log(`Stored cursor position for id ${id}: ${from}`); // Debug statement
+              // console.log(`Stored cursor position for id ${id}: ${from}`); // Debug statement
             } else {
               console.warn(`Invalid cursor position for id ${id}: ${from}`); // Debug statement
             }
--- a/src/client/CodeEditor/json1PresenceDisplay.ts
+++ b/src/client/CodeEditor/json1PresenceDisplay.ts
@@ -138,35 +138,13 @@ export const json1PresenceDisplay = ({
       }
             // Method to scroll the view to keep the cursor in view
             scrollToCursor(view) {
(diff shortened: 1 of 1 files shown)


DaCatDude:
The next goal here would be to add some sort of toggle, so that further work doesnt get confusing

src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -36,8 +40,14 @@ export const json1PresenceDisplay = ({
...
         const presenceState = {};
 
+        // Add the scroll event listener
+        //This runs for the arrow key scrolling, it should result in the users scrolling to eachother's location.
+        view.dom.addEventListener('scroll', () => {
+          this.scrollToCursor(view);
+        });
         // Receive remote presence changes.
         docPresence.on('receive', (id, presence) => {
+          console.log(`Received presence for id ${id}`, presence); // Debug statement
```
curran, line 50:
```suggestion
          // console.log(`Received presence for id ${id}`, presence); // Debug statement
```

src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -97,6 +107,12 @@ export const json1PresenceDisplay = ({
                 }),
               });
             }
+            if (view.state.doc.length >= from) { // Ensure position is valid
+              this.cursorPosition[id] = from; // Store the cursor position, important to run if we cant get the regular scroll to work
+              console.log(`Stored cursor position for id ${id}: ${from}`); // Debug statement
```
curran, line 112:
```suggestion
              // console.log(`Stored cursor position for id ${id}: ${from}`); // Debug statement
```

src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -97,6 +107,12 @@ export const json1PresenceDisplay = ({
...
+            } else {
+              console.warn(`Invalid cursor position for id ${id}: ${from}`); // Debug statement
```
curran, line 114:
```suggestion
              // console.warn(`Invalid cursor position for id ${id}: ${from}`); // Debug statement
```

curran (approved):
LGTM!

PR #719, opened 2024-06-08
fixed the wrong port opening in development mode

michaelhelper:
I added a new environment variable for the editor port to be 5173. The index.js file then checks the editor port variable and defaults to the current port if one is not specified.

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -310,9 +310,12 @@ server.listen(port, async () => {
       open(url);
     })();
   } else {
+    // Sets the port to the one specified in the environment 
+    // variable (for development) or the default port.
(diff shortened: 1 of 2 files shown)


src/server/index.js
```diff
@@ -310,9 +310,12 @@ server.listen(port, async () => {
       open(url);
     })();
   } else {
+    // Sets the port to the one specified in the environment 
+    // variable (for development) or the default port.
+    let livePort = process.env.EDITOR_PORT || port;
```
curran, line 315:
```suggestion
    const livePort = process.env.EDITOR_PORT || port;
```

curran (approved):
Excellent! Thank you so much.

PR #721, opened 2024-06-11
Focus ability

philippark:
Focus key for code editor
sidebar file navigation


curran (changes requested):
I think the package we would want to try is actually `react-aria`, not `react-aria-components`, because our goal is to make our existing DOM more accessible using the keyboard. We are already using Bootstrap as a a component library, so I'd rather not adopt a whole other component library.

Instead of trying to use the `ListBox` component, try using the custom hooks from `react-aria` that hook into the divs we already have.

I would suggest to try using `useListBox`, documented here https://react-spectrum.adobe.com/react-aria/useListBox.html

curran:
Closing in favor of https://github.com/vizhub-core/vzcode/pull/737

PR #727, opened 2024-06-11
changed colors of vizhub to pastel

michaelhelper:
Changed colors of vizhub to pastel theme might need to lighten the colors more.

image


michaelhelper:
Turns out, the package has seven layers of brackets, which can only be changed by copying the code and not utilizing the package.

curran (approved):
Very nice! This looks way better.

That's unfortunate about the package. Maybe we could copy in the entire library and modify it, it's so small anyway. Or we could fork the package, or submit PRs into that package.

Issue #726, opened 2024-06-11
Jump to definition: Get it to work for imports


curran:
Jump to definition would ideally also work for imports (specifically named imports).

Example:

```js
// You can import API functions like this from D3.js.
import { select } from 'd3';

// You can import local ES6 modules like this. See message.js!
import { message } from './myMessage';

// This line uses D3 to set the text of the message div.
select('#message').text(message);

```

I should be able to click on `select` on the last line, and have jump to the import.

Possible outcomes:

 * Ideally (stretch goal) it would open the definition of the exported variable in the other file (e.g. `import { foo } from './foo';` - this would only work for local files
 * Initially (easier goal) - it should just jump to the import line (this is all we can do for packages e.g. `d3` at the moment)

curran:
Closed in https://github.com/vizhub-core/vzcode/pull/733

Issue #725, opened 2024-06-11
Jump to definition: CTRL+Hover Styling 


curran:
As a person editing code with VZCode, I want it to be clear that I can CTRL+Click on a variable to go to its definition.

VSCode does:

 * `cursor: pointer`
 * Underline

Issue #724, opened 2024-06-11
Add CodeMirror Docs Reference Links re: Go to definition


curran:
Following https://github.com/vizhub-core/vzcode/pull/717#pullrequestreview-2111195361 , let's add URLs in comments for related documentation.

curran:
Closed in https://github.com/vizhub-core/vzcode/pull/733

PR #733, opened 2024-06-13
Jump to definition through CTRL + Hover + Click

jeffrey-asm:
The following will allow the current jump to definition to be based on the current mouse position in the editor DOM. When a user holds CTRL, an event listener will be dispatched to search for a potential defining location in the syntax tree for the identifier at the mouse location. If found, the identifier will be underlined and show a cursor on hover, indicating that a click on the identifier will move the view into the defining location. Note that this does not work for the actual defining location and only further usages in the editor.

This fixes the following issues

- #725 
- #726 
- #724


curran (approved):
LGTM!

PR #740, opened 2024-06-14
Second fix for back button

d-nieto246:
The changes to this code alone will open the files in VZCode, but also sync with the URL once you hit the back button to go back to a file you clicked on.

What you CAN do this moment: you can click the back button to move back to a file you opened, and it will go back to the file you worked on previously but you can only do it once, as of right now.

PR #746, opened 2024-06-14
Second fix for back button

d-nieto246:
The same pull request as https://github.com/vizhub-core/vzcode/pull/740 but in a new branch called "new-fix-back-button" so everything merged into main works along with my changes.

### DIFF
--- a/src/client/useActions.ts
+++ b/src/client/useActions.ts
@@ -18,14 +18,6 @@ export const useActions = (
     [dispatch],
   );
 
-  const setActiveFileLeft = useCallback(() => {
-    dispatch({ type: 'set_active_file_left' });
-  }, [dispatch]);
-
-  const setActiveFileRight = useCallback(() => {
-    dispatch({ type: 'set_active_file_right' });
-  }, [dispatch]);
-
   const openTab = useCallback(
     (tabState: TabState): void => {
       dispatch({
@@ -57,7 +49,6 @@ export const useActions = (
     [dispatch],
   );
 
-  // True to show the settings modal.
   const setIsSettingsOpen = useCallback(
     (value: boolean) => {
       dispatch({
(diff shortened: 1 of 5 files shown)


src/client/useURLSync.ts
```diff
@@ -10,65 +10,47 @@ import {
...
-    tabList.forEach(openTab);
-    setActiveFileId(activeFileId);
-  }, [
-    content,
-    searchParams,
-    setActiveFileId,
-    openTab,
-    isInitialized,
-  ]);
+    decodedTabList.forEach(openTab);
```
curran, line 51: We also need to close the tabs that were closed.

curran (changes requested):
Almost there! I noticed that it loses the "transient tabs" feature, which is where the tab opens and is italic. After these changes, when you open a tab, it is not transient.

We also need to close the tabs that are removed from the URL.

Instead of this

```js
decodedTabList.forEach(openTab);
```

I wonder if we could do something more straightforward like this:

```js
setTabList(decodedTabList)
```

There is currently no such thing as `setTabList`, but you could define it in `useActions`. 

You could even go one step further and define something in `useActions` that will update the tab list AND the active file in a single state update, which might be the cleanest solution. Maybe something like `setTabListAndActiveFileId({tabList, activeFileId})`.

src/client/vzReducer/index.ts
```diff
@@ -12,6 +12,7 @@ import { setIsDocOpenReducer } from './setIsDocOpenReducer';
 import { setThemeReducer } from './setThemeReducer';
 import { editorNoLongerWantsFocusReducer } from './editorNoLongerWantsFocusReducer';
 import { setUsernameReducer } from './setUsernameReducer';
+import { setTabListReducer } from './setTabListReducer'; // Imported a new reducer
```
curran, line 15:
```suggestion
import { setTabListReducer } from './setTabListReducer';
```

src/client/useActions.ts
```diff
@@ -102,18 +79,27 @@ export const useActions = (
...
+        type: 'set_tab_list',
+        tabList,
+      });
+    },
+    [dispatch],
+  );
+
   return {
     setActiveFileId,
-    setActiveFileLeft,
```
curran, line 107: These are used in the keyboard shortcuts. Deleting them will break things.

curran (changes requested):
When I run this branch, I see the following error:

image

Do you not see that when you run the code?

src/client/useActions.ts
curran: Before deleting things, it's a good idea to search for them in the codebase to see if they are used anywhere.

PR #750, opened 2024-06-18
Sidebarnav

philippark:
Scrolling sidebar files with keyboard


package.json
```diff
@@ -109,6 +109,8 @@
     "@codemirror/theme-one-dark": "^6.1.2",
     "@codemirror/view": "^6.28.1",
     "@lezer/highlight": "^1.2.0",
+    "@react-aria/listbox": "^3.12.1",
```
curran, line 112: Excellent choices!

src/client/VZSidebar/styles.scss
```diff
@@ -1,4 +1,56 @@
...
+    list-style: none;
+  }
+  
+  .list li {
+    outline: none;
+    cursor: default;
+  }
+  
+  .list li:nth-child(2n) {
+    background: var(--spectrum-alias-highlight-hover);
```
curran, line 12: Where is the `--spectrum-alias-highlight-hover` definition coming from? Did you verify that this resolves correctly? We may need a CSS import to make this work.

src/client/VZSidebar/index.tsx
```diff
@@ -94,6 +100,48 @@ export const VZSidebar = ({
     handleDrop,
   } = useDragAndDrop();
 
+  function List(props) {
+    let state = useListState(props);
```
curran, line 104: The style of this codebase prefers `const` over `let`. Please prefer `const` where possible. Thanks!

curran, line 104:
```suggestion
    const state = useListState(props);
```

src/client/VZSidebar/index.tsx
```diff
@@ -94,6 +100,48 @@ export const VZSidebar = ({
...
+    let ref = useRef();
```
curran, line 105:
```suggestion
    const ref = useRef();
```

src/client/VZSidebar/index.tsx
```diff
@@ -94,6 +100,48 @@ export const VZSidebar = ({
...
+    let { gridProps } = useGridList(props, state, ref);
```
curran, line 106:
```suggestion
    const { gridProps } = useGridList(props, state, ref);
```

src/client/VZSidebar/index.tsx
```diff
@@ -94,6 +100,48 @@ export const VZSidebar = ({
...
+      <ul {...gridProps} ref={ref} className="list">
+        {[...state.collection].map((item) => (
+          <ListItem key={item.key} item={item} state={state} />
+        ))}
+      </ul>
+    );
+  }
+
+  function ListItem({ item, state }) {
+    let ref = useRef(null);
```
curran, line 118:
```suggestion
    const ref = useRef(null);
```

src/client/VZSidebar/index.tsx
```diff
@@ -94,6 +100,48 @@ export const VZSidebar = ({
...
+    let { rowProps, gridCellProps, isPressed } = useGridListItem(
```
curran, line 119:
```suggestion
    const { rowProps, gridCellProps, isPressed } = useGridListItem(
```

src/client/VZSidebar/index.tsx
```diff
@@ -94,6 +100,48 @@ export const VZSidebar = ({
...
+      { node: item },
+      state,
+      ref
+    );
+  
+    let { isFocusVisible, focusProps } = useFocusRing();
```
curran, line 125:
```suggestion
    const { isFocusVisible, focusProps } = useFocusRing();
```

src/client/VZSidebar/index.tsx
```diff
@@ -94,6 +100,48 @@ export const VZSidebar = ({
...
+    let showCheckbox = state.selectionManager.selectionMode !== 'none' &&
```
curran, line 126:
```suggestion
    const showCheckbox = state.selectionManager.selectionMode !== 'none' &&
```

src/client/VZSidebar/index.tsx
```diff
@@ -199,11 +247,63 @@ export const VZSidebar = ({
             </div>
           </div>
         ) : filesExist ? (
+
+          <List
+            aria-label="Example List"
+            selectionMode="multiple"
+            selectionBehavior="replace"
+            onAction={(key) => handleFileDoubleClick(key)}
```
curran, line 255: I think this one can be simplified.
```suggestion
            onAction={handleFileDoubleClick}
```

curran (changes requested):
I left a review of the details. Overall nice work!

When I test it, here's what I see:

image

image

We seem to have lost the 100% width styling on the sidebar entries.

image

Also, there is a strange purple highlight over what looks like a parent div when I click. That does not look that good. Perhaps it could be made to match the border radius of the entries.

Also, how do I navigate through the files with the keyboard? I found that if I click on one, then use the arrow keys, it does not work as expected. I would expect the arrow keys to select between files, but it does not. I'm curious what interactions you are using to test this. Thanks!

PR #751, opened 2024-06-18
Added pin to UI for toggling auto follow

DaCatDude:
Auto follow pin added, fully functional.

Text changing based on the current state, toggle/untoggle. It might be worth changing the icon depending on the state too (pressed in pin vs non pressed, or maybe using something like this: 
https://www.svgrepo.com/svg/104431/man-silhouette-running-escaping 

instead of a pin. Let me know if you want that implemented.


src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -12,8 +12,10 @@ const debug = false;
 
 // TODO: Make this a setting in the UI.
 // See https://github.com/vizhub-core/vzcode/issues/739
-const enableAutoFollow = false;
-
+export let enableAutoFollow = false;
+export const toggleAutoFollowButton = () => {
+  enableAutoFollow = !enableAutoFollow;
```
curran, line 17: This is sort of an antipattern, but I get what is happening here. Perhaps we could replace this with a React Ref, which would be more idiomatic.

src/client/VZSidebar/index.tsx
```diff
@@ -233,6 +252,25 @@ export const VZSidebar = ({
...
+            overlay={
+              <Tooltip id="toggle-auto-follow">
+                {enableAutoFollow
+                  ? 'Disable Auto Follow'
+                  : 'Enable Auto Follow'}
+              </Tooltip>
+            }
+          >
+            <i
+              onClick={handleToggleFollow} //not implemented yet, replace with correct handle
```
curran, line 268:
```suggestion
              onClick={handleToggleFollow}
```

curran (approved):
Looks generally good, but the way `enableAutoFollow` is a mutable singleton is not ideal. I'll merge this as-is, but let's do a follow-up PR where `enableAutoFollow` is re-worked to be a `useState` plus `useRef` combo, something like this:

```js
const [enableAutoFollow, setEnableAutoFollow] = useState(false);
const enableAutoFollowRef = useRef(enableAutoFollow);
useEffect(() => {
  enableAutoFollowRef.current = enableAutoFollow;
}, [enableAutoFollow]);
```

Once this is set up, we can potentially move these definitions into a context that is visible to all the places that need to know about `enableAutoFollow`.

PR #753, opened 2024-06-21
729  automatically open the file that the followed person navigates to by either opening a new tab or switching to an already open tab

DaCatDude:
Working on the basic framework to impliment this. I am currently stuck on how openTab works. The react hooks give an error, so this is still a work in progress.

### DIFF
--- a/src/client/CodeEditor/json1PresenceDisplay.ts
+++ b/src/client/CodeEditor/json1PresenceDisplay.ts
@@ -12,8 +12,10 @@ const debug = false;
 
 // TODO: Make this a setting in the UI.
 // See https://github.com/vizhub-core/vzcode/issues/739
-const enableAutoFollow = false;
-
+export let enableAutoFollow = false;
+export const toggleAutoFollowButton = () => {
+  enableAutoFollow = !enableAutoFollow;
+};
 // Deals with receiving the broadcasted presence cursor locations
 // from other clients and displaying them.
 //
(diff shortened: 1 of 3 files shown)


src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -6,14 +6,20 @@ import {
...
+import { useActions } from '../useActions';
 
 const debug = false;
 
 // TODO: Make this a setting in the UI.
 // See https://github.com/vizhub-core/vzcode/issues/739
-const enableAutoFollow = false;
-
+export let enableAutoFollow = true;
+export const toggleAutoFollowButton = () => {
```
curran, line 20: I changed all this around in https://github.com/vizhub-core/vzcode/pull/769

It may be best to try building this feature fresh from `main` at this point.

src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -44,7 +50,7 @@ export const json1PresenceDisplay = ({
         //  * Keys are presence ids
         //  * Values are presence objects as defined by ot-json1-presence
         const presenceState = {};
-
+        const  { openTab } = useContext ( VZCodeContext );
```
curran, line 53: Close but no cigar! `useContext` will not work here because we're not inside a React function component. You can try passing in `openTab` as an option to ` json1PresenceDisplay` instead, which should work. Good luck!

curran (changes requested):
Excellent attempt! It seems doable with the suggestions I made in the comments:

 * Branch fresh off the latest `main`
 * Add `openTab` as an option to `json1PresenceDisplay`, then pass it in from the outside

Good luck!

curran:
See also https://github.com/vizhub-core/vzcode/pull/781

PR #777, opened 2024-07-09
Search: Auto-focus the input field

jeffrey-asm:
The following is a small change that ensures that whenever the search view is mounted, the search input will auto-focus, improving user experience.


curran (approved):
Excellent! Thank you.

PR #778, opened 2024-07-09
Updated keyboard docs to include Alt-1 and Alt-2

philippark:


curran:
I'll hold off on merging this one until these are fully working.

We could potentially change the keyboard shortcuts if Chrome already uses Alt+{number} for tab navigation.

PR #780, opened 2024-07-09
Display New File Icons

jeffrey-asm:
The following adds the newly added file icons based on file extensions from the search feature into the Files view. Icons are updated on file rename, and an unused file icon was removed, which was introduced in the same search feature PR.


curran (approved):
LGTM!

Issue #658, opened 2024-04-12
Bug: Renaming files and directories has no name validation


enuessle:
File name validation is only checked on creation of new files. Renaming existing files in VZcode from the sidebar gets around all name validation.

![File Renaming](https://github.com/vizhub-core/vzcode/assets/77037513/7931a633-4770-471b-8805-1dadf553acc1)

curran:
Really great idea! I'm all for it.

PR #786, opened 2024-07-11
Fix Renaming Bug

philippark:


philippark:
Next on the agenda is to make the styling better looking

src/client/VZSidebar/Item.tsx
```diff
@@ -64,6 +64,19 @@ export const Item = ({
     [id, setHoveredItemId],
   );
 
+  const validateFileName = useCallback(
```
curran, line 67: Please refactor this so that there is only a single definition of `validateFileName` in the codebase. I see there is another definition of this over in `src/client/VZSidebar/CreateDirModal.tsx`. Please pull this function out into its own separate file, then call it from both places. Thanks!

curran (changes requested):
Thank you for this! I left a comment about a suggested refactoring.

src/client/VZSidebar/Item.tsx
curran: Something like this:

```
export const validateFileName = ({files}) => {
  // Returns true if file name is valid, false otherwise.
  return useCallback(
    (fileName: string) => {
      let valid;
      // General Character Check
      const regex =
        /^[a-zA-Z0-9](?:[a-zA-Z0-9 ./+=_-]*[a-zA-Z0-9])?$/;
      valid = regex.test(fileName);

      // Check for Duplicate Filename
      for (const key in files) {
        if (fileName + '/' === files[key].name) {
          valid = false;
        }
      }

      return valid;
    },
    [files],
  );
}
```

This is a "custom hook" in React parlance.

curran:
I'll take it from here. Thanks for this work!

curran:
Does this close a specific existing issue? If so which one?

philippark:
Hi, yes it resolves this: https://github.com/vizhub-core/vzcode/issues/658

Issue #790, opened 2024-07-12
Reorganize keyboard shortcuts documentation for sidebar buttons


curran:
![image](https://github.com/user-attachments/assets/a48cae12-6ab4-4c14-8299-0500b06a16f0)

This is a bit hard to read through as it's a paragraph format. Let's clean it up to be nested bullets, or something similar.

Issue #789, opened 2024-07-12
Keyboard shortcuts in tooltips for sidebar buttons


curran:
Ideally we'd explain the keyboard shortcuts on hover in the sidebar icons, similar to the run icon

![image](https://github.com/user-attachments/assets/11c75184-af56-46f7-85b9-993810633426)

PR #795, opened 2024-07-16
Update Sidebar Button Tooltips

jeffrey-asm:
The following updates each sidebar button tooltip to display their respective keyboard shortcut. An additional shortcut (`Ctrl +,`) to trigger the settings view was also added, which is based on the VS Code implementation. This fixes [#789](https://github.com/vizhub-core/vzcode/issues/789) and potentially removes the need for [#790](https://github.com/vizhub-core/vzcode/issues/790) due to the complexity of total shortcuts to represent.

### DIFF
--- a/src/client/useKeyboardShortcuts.ts
+++ b/src/client/useKeyboardShortcuts.ts
@@ -218,8 +218,10 @@ export const useKeyboardShortcuts = ({
       if (event.ctrlKey && event.shiftKey) {
         // Handle keyboard shortcuts related to the side bar icons
         document.getElementById(sideBarKeyBoardMap[event.key])?.click();
-      }
-
+      } else if (event.ctrlKey && event.key === ',') {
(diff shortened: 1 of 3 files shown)


src/client/useKeyboardShortcuts.ts
```diff
@@ -218,8 +218,10 @@ export const useKeyboardShortcuts = ({
       if (event.ctrlKey && event.shiftKey) {
         // Handle keyboard shortcuts related to the side bar icons
         document.getElementById(sideBarKeyBoardMap[event.key])?.click();
-      }
-
+      } else if (event.ctrlKey && event.key === ',') {
```
curran, line 221:
```suggestion
      // Mimics VSCode shortcut to opn the settings
      } else if (event.ctrlKey && event.key === ',') {
```

curran (approved):
LGTM!

PR #796, opened 2024-07-16
finally fixed

michaelhelper:
Added a cross-env package for windows-Unix cross compatibility for environment variables and changed the environment variable assignment accordingly.


curran (approved):
LGTM!

PR #801, opened 2024-07-19
Context menu

philippark:


philippark:
yet to do:
- actual interactive buttons on the context menu
- halt everything on the screen until menu is unclicked.

package.json
```diff
@@ -148,6 +148,7 @@
     "react-router-dom": "^6.24.1",
     "sharedb": "^5.0.2",
     "sharedb-client-browser": "^5.0.2",
+    "styled-components": "^6.1.12",
```
curran, line 151: Let's not adopt `styled-components` please. We are already all set up with Sass.

curran (changes requested):
Please use regular CSS/SaSS. Thanks!

PR #811, opened 2024-07-26
Search Results Keyboard Navigation

jeffrey-asm:
The following adds a few keyboard navigation options to traverse through a given search result batch of files and matching occurrences within that given file. Once a user hits `Tab` on the search input container, keyboard navigation can be utilized. The UI was also changed to display the close icon on the current file or line occurrence if it is the active element. Future work could involve changing the `Tab` behavior to toggle the close icon, but `Enter / Space` handlers will need to be accounted for to close the result.

- `Tab` - Focus relative file
- `Arrow-up / Arrow-down` - navigate upward or downward
- `Arrow-right` - Enter search file tree
- `Arrow-left` - Focus relative file or flatten a result file
- `Enter / Space` - Jump to the relative file or occurrence of the pattern


curran:
Notes from QA:

 * The layout seems to jump around as you navigate. Could be to do with the borders. Also the X button seems to cause the layout to shift. Solutions could be: maybe try "outline" instead of border, or investigate "box sizing" and CSS properties around border. With the X, we could try `position: absolute`, which requires a parent be `position: relative`.
 * There is no hover state on the X button. Maybe we could make it similar to the X hover state on the tabs.

src/client/VZSidebar/Search.tsx
```diff
@@ -85,39 +226,38 @@ export const Search = () => {
...
       </Form.Group>
       {Object.keys(results).length >= 1 &&
-      pattern.trim().length >= 1 ? (
-        <div className="search-results">
-          {Object.entries(results)
-            .filter(
-              ([_, file]) => file.visibility !== 'closed',
-            )
-            .map(([fileId, file]: [string, SearchFile]) => (
+        pattern.trim().length >= 1 ? (
```
curran, line 235:
```suggestion
        !isSearchQueryEmpty ? (
```

then something like

```
const = isSearchQueryEmpty = useMemo(() =>  pattern.trim().length >= 1)
```

src/client/VZSidebar/Search.tsx
```diff
@@ -85,39 +226,38 @@ export const Search = () => {
...
+        <div
+          onKeyDown={handleKeyDown}
+          tabIndex={0}
+          className="search-results">
+          {files.map(([fileId, file]: [string, SearchFile], index) => {
+            const matches = file.matches;
+            return (
               <div
                 className="search-result"
```
curran, line 244: This could be a component, e.g. `SearchResult`.

src/client/VZSidebar/Search.tsx
```diff
@@ -85,39 +226,38 @@ export const Search = () => {
...
                 key={file.name}
               >
-                <div className="search-file-heading">
+                <div
+                  onClick={() => focusFileElement(fileId, index)}
```
curran, line 248: Ideally `useCallback` for this

src/client/VZSidebar/Search.tsx
```diff
@@ -85,39 +226,38 @@ export const Search = () => {
...
+                  id={fileId}
+                  className={`search-file-heading 
+                      ${(focusedIndex == index && focusedChildIndex == null)
+                      ? 'active' : ''}`}
+                >
                   <div className="search-file-title">
                     <div
                       className="arrow-wrapper"
+                      onClick={() => flattenResult(fileId, file)}
```
curran, line 257: Ideally `useCallback` for this

src/client/VZSidebar/Search.tsx
```diff
@@ -127,26 +267,32 @@ export const Search = () => {
...
-                        );
-                      }}
-                    >
-                      <CloseSVG />
-                    </div>
+                    {
+                      (index == focusedIndex && focusedChildIndex == null) ? (
+                        <span
+                          className="search-file-close"
+                          onClick={(event) => {
```
curran, line 274: Ideally `useCallback`.

src/client/VZSidebar/Search.tsx
```diff
@@ -159,48 +305,68 @@ export const Search = () => {
                         match.index + pattern.length,
                       );
 
+                      const identifier = file.name + "-" + match.line;
```
curran, line 308: Ideally we shouldn't need this if we are using `useRef` to get at the DOM elements.

src/client/VZSidebar/Search.tsx
```diff
@@ -159,48 +305,68 @@ export const Search = () => {
...
+                              ${(focusedIndex == index && focusedChildIndex == childIndex)
+                              ? 'active' : ''}`}
                         >
-                          {before}
-                          <span className="search-pattern">
-                            {hit}
-                          </span>
-                          {after}
-                        </p>
+                          <p
```
curran, line 319:
```suggestion
                          <div
```

src/client/VZSidebar/Search.tsx
```diff
@@ -159,48 +305,68 @@ export const Search = () => {
...
+                            key={
+                              file.name +
+                              ' - ' +
+                              match.line +
+                              ' - ' +
+                              match.index
+                            }
+                            onClick={() => {
```
curran, line 327: Ideally `useCallback` here

src/client/VZSidebar/Search.tsx
```diff
@@ -159,48 +305,68 @@ export const Search = () => {
...
+                              setSearchFocusedIndex(index, childIndex);
+
+                              if (editorCache.get(fileId)) {
```
curran, line 330: Needs to use the new key pattern (could be done later)

src/client/VZSidebar/Search.tsx
```diff
@@ -159,48 +305,68 @@ export const Search = () => {
...
+                            <span className="search-pattern">
+                              {hit}
+                            </span>
+                            {after}
+                          </p>
+                          {
+                            (focusedIndex == index && focusedChildIndex === childIndex) && (
+                              <span
+                                className="search-file-close"
+                                onClick={(event) => {
```
curran, line 346: Could be `useCallback`

src/client/useActions.ts
```diff
@@ -114,6 +114,35 @@ export const useActions = (
...
+    ) => {
+      dispatch({
+        type: 'hide_search_results_line',
+        files: files,
+        id: id,
+        line: line
+      });
+    },
+    [dispatch],
+  );;
```
curran, line 132:
```suggestion
  );
```

curran (approved):
LGTM!!!!

PR #815, opened 2024-07-30
presence color revamp and intial pass at syncing

DaCatDude:
This is an intail pass at color matching. I tried revamping presence to store a color componant but im running into an issue that I dont know how to solve.

Console log:
image

in index.tsx, you will see the console logs before and after I try to access and assign the userColor via update.userColor (the same way the username is assigned).

However, for some reason, the colors end up as null, even though the update object is clearly logged to have a populated color component. 

I tried to fix this but I have no idea how, as after those are recorded as null, the next console log shows that update still has its userColor component.

### DIFF
--- a/src/client/CodeEditor/json1PresenceDisplay.ts
+++ b/src/client/CodeEditor/json1PresenceDisplay.ts
@@ -88,14 +88,11 @@ export const json1PresenceDisplay = ({
               presence,
             );
 
-// Initialize or update presence with user color
-const initializePresence = (id: string, presence: Presence): Presence => {
-  return assignUserColor(presence, id);
-};
+
 
   // If the presence is in the current file, update the presence state with the color.
   if (isPresenceInCurrentFile) {
-    presenceState[id] = initializePresence(id, presence);    
+    presenceState[id] = presence;    
     console.log('Structure of original presence object:', JSON.stringify(presenceState[id], null, 2));
   } else if (presence) {
     // Otherwise, delete the presence state.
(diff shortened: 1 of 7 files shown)


DaCatDude:
when  I add the console log below
        console.log('Structure of update object:', JSON.stringify(update, null, 2));
        
        This is what is logged overall:
        
image

even though userColor clearly exists, it doesnt show as part of the stringify

curran:
I think we can compute the color in the same way it's computed elsewhere, which I believe is based on the name.

curran:
Probably we can just extract this into a function and call it from both places:

```js
              const userColor = new ColorHash({
                lightness: 0.75,
              }).rgb(id);
```

from `src/client/CodeEditor/json1PresenceDisplay.ts`

DaCatDude:
Added a function, now the two objects sync, but the background color still doesnt.
image

In the editor, the background is transparent, because 
              style={{ backgroundColor: indicator.userColor }}
indicator.userColor returns a null color value for some reason.

curran:
@DaCatDude Thanks for the update! I would suggest not to add a `indicator.userColor` property, but instead just compute the color when  you need it, something like `indicatorColor(indicator)`.

src/client/CodeEditor/json1PresenceDisplay.ts
```diff
@@ -88,14 +88,15 @@ export const json1PresenceDisplay = ({
               presence,
             );
 
-            // Generate user color based on the id
-  const userColor = new ColorHash({ lightness: 0.75 }).rgb(id).join(',');
+// Initialize or update presence with user color
+const initializePresence = (id: string, presence: Presence): Presence => {
+  return assignUserColor(presence, id);
```
curran, line 93: This may not work, the presence object may be re-initialized or something, as this is managed by ShareDB. Better not try to modify it.

src/client/presenceColor.ts
```diff
@@ -0,0 +1,7 @@
+import ColorHash from 'color-hash';
+import { Presence } from '../types';
+
+export const assignUserColor = (presence: Presence, id: string): Presence => {
```
curran, line 4: The concept is almost correct - I would suggest to have this just return the color, rather than mutating the presence object.

PR #829, opened 2024-08-09
split pane changes

michaelhelper:


curran:
Closing in favor of https://github.com/vizhub-core/vzcode/pull/831

Some good ideas here! Specifically

```
for (const child of activePane.children) {
```
We'll need this sort of logic in the new `SplitPaneView` component! More like `activePane.children.map(child => <SplitPaneView/>)` or something - a recursive React component, which is super cool.

Also the

```
 // TODO Copy the tab list from the active pane
```

is totally valid and still needs to be done. Cheers!

Issue #705, opened 2024-06-05
Split Pane View


curran:
As a VZCode user editing two files at once, I want to be able to see them side by side, so that I can easily switch, or copy-paste between them.

michaelhelper:
https://codemirror.net/examples/split/

curran:
![image](https://github.com/user-attachments/assets/d1bf4e8e-e0da-4e4e-8f74-af31d1531794)

curran:
![image](https://github.com/user-attachments/assets/af5ae5e6-3e45-43c2-b1a5-4aa89684f41a)

curran:
![image](https://github.com/user-attachments/assets/cd7f5f57-ada1-4e81-a1fe-52cf2c7792ee)

curran:
Data structure ideas for changes to `VZState`

```
// The leaf node of the tree data structure
type LeafPane = {
  type: 'pane';
  // The list of tabs
  // Mutually exclusive with `children`.
  tabList: Array<TabState>;
};

// Internal node of the tree data structure
type SplitPane = {
  type: 'splitPane';
  // Which orientation is it? Vertical split or horizontal split?
  // Applies only to `children`
  orientation: 'vertical' | 'horizontal';

  // The children panels
  // Mutually exclusive with `tabList`.
  children: Array<Pane>;
};

// The node data structure of the split pane tree
type Pane = LeafPane | SplitPane;
```

curran:
```js
// The leaf node of the tree data structure
type LeafPane = {
  type: 'leafPane';
  // The list of tabs
  // Mutually exclusive with `children`.
  tabList: Array<TabState>;
};

// Internal node of the tree data structure
type SplitPane = {
  type: 'splitPane';
  // Which orientation is it? Vertical split or horizontal split?
  // Applies only to `children`
  orientation: 'vertical' | 'horizontal';

  // The children panels
  // Mutually exclusive with `tabList`.
  children: Array<Pane>;
};

// The node data structure of the split pane tree
type Pane = LeafPane | SplitPane;

// The shape of the state managed by the reducer.
export type VZState = {
  // The state of the split pane (tree data structure).
  pane: Pane;
```

curran:
Path forward:

 * Add the icon
 * When clicking the icon, split the pane by dispatching an action
 * In the reducer that handles that action, 1.) isolate the pane that you want to split (at first this can be hardcoded to `root`), or code it to split the currently active pane.
 
 
[...]

PR #831, opened 2024-08-09
Progress towards split pane

curran:
Towards #705


curran:
Supercedes https://github.com/vizhub-core/vzcode/pull/829/files , which has some nice related ideas

PR #838, opened 2024-08-13
worked on split pane view

michaelhelper:


src/client/VZMiddle.tsx
```diff
@@ -67,6 +50,99 @@ const PaneView = ({
...
+  customInteractRules,
+  allowGlobals,
+  enableAIAssist,
+  aiAssistEndpoint,
+  aiAssistOptions,
+  aiAssistTooltipText,
+  aiAssistClickOverride,
+}) => {
+  console.log('SplitPaneView');
+  return (
```
curran, line 65:
```suggestion
  return pane.children.map((pane:Pane) => { ... });
```

curran (commented):
Thanks for sharing WIP! Added an idea for using `.map` to iterate over the panes.

PR #865, opened 2024-10-03
"Copy all files" text

alizavirji:


curran:
Something that would be really cool to do with this: we can make it copy all the files in a format that LLMs can understand! That would be super useful.

I started on a related small package that does parsing and formatting for LLMs:

https://github.com/curran/llm-code-format

curran:
This actually would be really nice to have. Reopening the PR, I'll try to finish it.

Issue #722, opened 2024-06-11
Ability to Toggle Rainbow Brackets on/off


curran:
As a VZCode user, I want to be able to turn off Rainbow Brackets if it's annoying.

curran:
This could go into the Settings modal

![image](https://github.com/vizhub-core/vzcode/assets/68416/75bd6cac-cb9a-4136-87d4-5a84ea2d70bb)

PR #869, opened 2024-10-11
Rainbow bracket toggle button

TroyG-11:
https://github.com/vizhub-core/vzcode/issues/722

Working on this right now, created a button so far to toggle.

### DIFF
--- a/src/client/CodeEditor/getOrCreateEditor.ts
+++ b/src/client/CodeEditor/getOrCreateEditor.ts
@@ -107,6 +107,13 @@ const getAtPath = (obj, path) => {
   return current;
 };
 
+// Extend the EditorCacheValue type to include the compartments and the updateRainbowBrackets method
+interface ExtendedEditorCacheValue extends EditorCacheValue {
+  themeCompartment: Compartment;
+  rainbowBracketsCompartment: Compartment;
+  updateRainbowBrackets: (enabled: boolean) => void;
+}
+
 // Gets or creates an `editorCache` entry for the given file id.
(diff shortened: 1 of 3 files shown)


TroyG-11:
No current functionality as of yet, just added the button dropdowns.

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -106,6 +107,13 @@ const getAtPath = (obj, path) => {
   return current;
 };
 
+// Extend the EditorCacheValue type to include the compartments and the updateRainbowBrackets method
```
curran, line 110: Looks great!

curran (approved):
This looks like the right way to do it! Thank you so much for this.

curran (changes requested):
I tested it and it seems like it does not work. Here's the screen recording of the test:


https://github.com/user-attachments/assets/c6b1a67c-a26d-4c1a-b448-c116016a3386

TroyG-11:
> I tested it and it seems like it does not work. Here's the screen recording of the test:
> 
>  rainbowBracketsTest.mp4

Hi, I've been working on this on a separate branch right now. I had the PR up as a draft since it isn't completed yet. Will be finishing it up soon.

TroyG-11:
Should be ready. 

It might have a quick delay before the render update on the rainbow brackets happen after switching toggle.

curran (approved):
LGTM! Tested locally and it works. There is a delay, it seems like maybe you need to interact with the editor for the change to take hold, but that's OK.

Thank you!

PR #872, opened 2024-10-15
Update to 4o

darrenW04:
I'm going to pr this for now, my implementation for claude was going to include this but that seems more indepth and would take more time. So I would rather have this pushed instead of waiting on those changes.


curran:
Sweeet! Thank you.

Issue #561, opened 2024-02-27
Saving / Saved Status


curran:
Following https://github.com/vizhub-core/vzcode/issues/456

As a user who just made a change, I want to be sure that the change was saved to the server.

```js
import { useState, useEffect, useCallback, useRef } from 'react';

import { runDelay } from '../constants';

export const usePending = (shareDBDoc) => {
  const lastOpTimerRef = useRef();
  const [pending, setPending] = useState(false);

  const resetOnNoPending = useCallback(() => {
    if (!shareDBDoc) return;

    shareDBDoc.whenNothingPending(() => {
      setPending(false);
    });
  }, [shareDBDoc, setPending]);

  useEffect(() => {
    if (!shareDBDoc) return;

    shareDBDoc.on('before op', () => {
      setPending(true);

      clearTimeout(lastOpTimerRef.current);

      lastOpTimerRef.current = setTimeout(() => {
        // should be triggered only once, cause upcoming ops unschedule previous resetOnNoPending
        resetOnNoPending();
      }, runDelay);
    });
  }, [shareDBDoc, resetOnNoPending]);

  return pending;
};
```

curran:
![image](https://github.com/vizhub-core/vzcode/assets/68416/0f527600-c142-46d6-8c35-0798edc36290)

![image](https://github.com/vizhub-core/vzcode/assets/68416/88c44b2a-f76c-4713-9925-5707990847d7)

curran:
Ideal behavior:

 * On page load, it should say "Connecting..." and be yellow
 * Then should transition to "Connected" and be green
 * If a remote user makes a change, the label and color should not change at all
 * If the user makes change, it should transition to "Saving..." and be yellow
 * After the change is saved, it should transition to "Saved." and be green
 * Then after some time (several seconds) should transition to "Connected" and be green

curran:
Current status: was partially implemented in

 * https://github.com/vizhub-core/vzcode/pull/874

I think we just need to bring back those changes to working order.

PR #874, opened 2024-10-19
561 saving/saved status

sxkgir:
A user who just made a change can see on the bottom left of the sidebar the status of it being saved or saving. When the user has stopped editing for around 2 seconds the status will be set to saved and back to connected. The new feature works fine on my end, let me know if there is any problems.


curran:
Closes #561

curran (approved):
LGTM! Tested locally, works great! Thank you so much.

PR #877, opened 2024-11-19
430 number dragger should work outside editor

IBurntMyBagel:
Current Behavior: Allows the user to drag the number outside the code editor tab. Displays a preview of the number value above the number being dragged.
Changed: No longer updates the value of the number being dragged in real time and now updates the value when user stops dragging number.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1788,15 +1788,26 @@
       }
     },
     "node_modules/acorn": {
-      "version": "8.11.3",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.11.3.tgz",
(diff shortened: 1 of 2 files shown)


curran (changes requested):
While this is a good idea in a way, it defeats the entire purpose of the interactive widgets, which is to trigger real-time updates to the code so that users can interactively tweak values, and that tweaking works well with hot reloading.

The requirement for the issue is to _preserve_ the real-time updates to the code, while also having it work outside the editor area. My apologies if this was not clear.

IBurntMyBagel:
Ah no worries. I understand the importance of keeping the updating in real time feature for the number dragger. Since it's currently able to be dragged off screen, I'll try and work on the issue more and get it to update while off screen as the next step.

Thanks for the feedback.

PR #878, opened 2024-11-19
189 color pick position edited

IBurntMyBagel:
When users press alt + left click on the color circle they will now have the color picker appear near their mouse instead of having the color picker remain in the top left of their screen. Users can click outside of the color picker to close the color picker.

### DIFF
--- a/src/client/CodeEditor/InteractiveWidgets/colorPicker.ts
+++ b/src/client/CodeEditor/InteractiveWidgets/colorPicker.ts
@@ -8,39 +8,94 @@ export const colorPickerRegex = /#[0-9A-Fa-f]{6}/g;
 // Works without quotes to support CSS.
 export const colorPicker = (
   onInteract?: () => void,
-): InteractRule => ({
-  regexp: colorPickerRegex,
-  cursor: 'pointer',
-  onClick(
-    text: string,
-    setText: (newText: string) => void,
-  ) {
-    const startingColor: string = text;
+):  InteractRule => {
+  // Create the color picker element when the page loads
+  const sel: HTMLInputElement = document.createElement('input');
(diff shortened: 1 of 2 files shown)


src/client/CodeEditor/InteractiveWidgets/colorPicker.ts
```diff
@@ -8,39 +8,135 @@ export const colorPickerRegex = /#[0-9A-Fa-f]{6}/g;
...
+  // Create the color picker element when the page loads
+  const sel: HTMLInputElement = document.createElement('input');
+  sel.type = 'color';
+  sel.style.position = 'fixed';
+  sel.style.left = '-9999px'; // Initially off-screen
+  sel.style.top = '-9999px';
+  document.body.appendChild(sel);
+
+  // Create a close button element to hide the color picker
+  const closeButton: HTMLButtonElement = document.createElement('button');
```
curran, line 21: Nice. There could be some additional work here to make it more similar to the "X" buttons in other parts of the UI, like closing tabs, but this is awesome as a first version. Excellent work! Thank you so much.

curran (approved):
LGTM!

image

src/client/CodeEditor/InteractiveWidgets/colorPicker.ts
curran: I was testing this more and there seems to be a regression actually, where the text is not getting updated.

I'm considering temporarily reverting these changes so the color picker edits the text again.

Thanks for your work on this!

curran: Fixed in https://github.com/vizhub-core/vzcode/pull/884

Issue #683, opened 2024-05-15
Custom styling of CodeMirror Search


curran:
This CodeMirror search UI uses the default styling. It feels out of place. It also conflicts with the AI Assist icon.

![image](https://github.com/vizhub-core/vzcode/assets/68416/bc19382c-eb35-4e5a-93ff-0b74bac4b641)

Ideally the style would be consistent with the rest of the app.

PR #880, opened 2024-11-26
Updated search UI

TroyG-11:
https://github.com/vizhub-core/vzcode/issues/683

Matches theme colors now.

Also updated styling and highlight colors.


curran (approved):
Nice! This is definitely getting closer.

Before:

image

After:

image

This is a great start! It sets up the space where we can work to iterate the styles and make them more like the rest of the app, which uses the Bootstrap styles for buttons etc.

PR #886, opened 2024-12-10
Expose system fonts

sxkgir:


src/client/VZSettings.tsx
```diff
@@ -8,17 +8,46 @@ import React, {
...
-import { toggleRainbowBrackets } from './CodeEditor/rainbowBrackets';
 
-const fontSizes = ['10px','12px','14px','16px', '18px', '20px', '24px'];
+// Font size options for the editor
+const fontSizes = ['10px', '12px', '14px', '16px', '18px', '20px', '24px'];
+
+// List of system fonts to be detected for availability
+const systemFonts = [
+  '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Helvetica Neue', 'Arial', 'sans-serif',
+  'Courier New', 'Georgia', 'Tahoma', 'Verdana', 'Times New Roman', 'Trebuchet MS', 'Comic Sans MS'
```
curran, line 18: These should focus on monospace fonts.

src/client/VZSettings.tsx
```diff
@@ -203,45 +174,36 @@ export const VZSettings = ({
           <select
             className="form-select"
             value={theme}
-            onChange={handleThemeChange}
+            onChange={(e) => setTheme(e.target.value as ThemeLabel)}
```
curran, line 177: The `useCalback` is a good optimization to keep in place. Otherwise performance may degrade if we do this all over the place (since removing and re-adding event listeners each render can be expensive of there are lots of these).

src/client/VZSettings.tsx
```diff
@@ -250,29 +212,12 @@ export const VZSettings = ({
...
-          </select>
-          <Form.Text className="text-muted">
-            Toggle Rainbow Brackets
-          </Form.Text>
-        </Form.Group>
-
       </Modal.Body>
       <Modal.Footer>
+        <Button variant="secondary" onClick={resetSettings}>
+          Reset to Default
```
curran, line 219: Very nice addition!

curran (approved):
Very nice! Thank you so much.

Issue #844, opened 2024-09-11
Adopt val-town/codemirror-ts for TypeScript support


curran:
It might be a Good Idea to replace all our home grown TypeScript linting and completions support for this solution which may be more robust.

https://github.com/val-town/codemirror-ts

curran:
Related https://github.com/val-town/codemirror-ts/issues/40

PR #913, opened 2025-04-08
Upgrade deps, Drop TypeScript worker support

curran:


benbenbenbenbenben:
Out of interest, why was this dropped @curran ?

curran:
@benbenbenbenbenben Thanks for asking! It was dropped because we want to make way for this other solution:

 * https://github.com/vizhub-core/vzcode/issues/844

The existing solution that was here was being overly aggressive across the board, and since we plan to add the other extension for TypeScript support, I figured I would just remove it. That said, if there's something you miss about it, I would be open to bringing it back, since the new solution has not been pulled in yet.

I suppose the main issue is that we were using the TypeScript language server for essentially non-TypeScript linting purposes, and we were getting all sorts of spurious warnings about types in `.js` and `.jsx` files. In an ideal world, we would have an ESLint-based linter CodeMirror extension running for `.js` and `.jsx` files, and use [val-town/codemirror-ts](https://github.com/val-town/codemirror-ts) for `.ts` and `.tsx` files.

I'm curious, is there something you miss from the old implementation that you found useful? Maybe we could bring it back.

benbenbenbenbenben:
I'm was looking for a browser operable type checker for typescript/js snippets, something with ATA, and I ended up here. I hadn't come across vzcode before a few days ago, but now that I have I'll keep an eye on things. I suppose I was asking really because I wondered what direction development was heading in.

Thank you for filling me in! val-town's package might be what I need too. I'll be back to look at vzcode when I've made some space for other projects!

PR #917, opened 2025-05-10
Integrate ESLint for enhanced code linting

fantong11:
Introduce ESLint support with asynchronous linting capabilities, enabling `.jsx` and `.js` linting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,7 +160,6 @@
     "sharedb-client-browser": "^5.1.1",
     "uuid": "^11.1.0",
     "vizhub-ui": "^3.23.0",
-    "vzcode": "file:",
     "ws": "^8.18.1"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "@codemirror/lang-javascript": "^6.2.3",
     "@codemirror/lang-json": "^6.0.1",
     "@codemirror/lang-markdown": "^6.3.2",
+    "@codemirror/lint": "^6.8.5",
     "@codemirror/state": "^6.5.2",
     "@codemirror/theme-one-dark": "^6.1.2",
     "@codemirror/view": "^6.36.5",
@@ -140,6 +141,8 @@
     "d3-array": "^3.2.4",
     "diff-match-patch": "^1.0.5",
     "dotenv": "^16.4.7",
+    "eslint": "^9.26.0",
+    "eslint-linter-browserify": "^9.26.0",
     "express": "^5.1.0",
     "ignore": "^7.0.3",
     "json0-ot-diff": "^1.1.2",
@@ -156,6 +159,7 @@
     "sharedb-client-browser": "^5.1.1",
     "uuid": "^11.1.0",
     "vizhub-ui": "^3.23.0",
+    "vzcode": "file:",
     "ws": "^8.18.1"
   },
   "devDependencies": {
@@ -173,6 +177,6 @@
   },
   "optionalDependencies": {
     "@rollup/rollup-darwin-arm64": "^4.39.0",
-    "@rollup/rollup-win32-x64-msvc": "^4.39.0"
+    "@rollup/rollup-win32-x64-msvc": "^4.40.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -176,7 +176,6 @@
     "vitest": "^3.1.1"
   },
   "optionalDependencies": {
-    "@rollup/rollup-darwin-arm64": "^4.39.0",
-    "@rollup/rollup-win32-x64-msvc": "^4.40.2"
+    "@rollup/rollup-darwin-arm64": "^4.39.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -144,6 +144,7 @@
     "eslint": "^9.26.0",
     "eslint-linter-browserify": "^9.26.0",
     "express": "^5.1.0",
+    "globals": "^16.1.0",
     "ignore": "^7.0.3",
     "json0-ot-diff": "^1.1.2",
     "livekit-server-sdk": "^2.11.0",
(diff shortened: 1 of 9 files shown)


src/client/useESLint/worker.ts
```diff
@@ -0,0 +1,120 @@
+import * as eslint from 'eslint-linter-browserify';
+import globals from 'globals';
```
curran, line 2: We might need an `npm install -S globals`, I don't think this is in the `package.json` dependencies.

src/client/useESLint/worker.ts
```diff
@@ -0,0 +1,120 @@
...
+    self.postMessage({
+      diagnostics: [],
+      requestId,
+      error: 'Invalid code received',
+    });
+    return;
+  }
+
+  // Only lint .js or .jsx files
+  if (!fileName || !/\.(js|jsx)$/i.test(fileName)) {
```
curran, line 70: Aha! So this is why the `fileName` is needed. I wonder if we could move this test to be outside of the Worker, where we just don't send the message at all if we are not in .js or .jsx files.

src/client/useESLint/worker.ts
```diff
@@ -0,0 +1,120 @@
...
+    self.postMessage({ diagnostics: [], requestId });
+    return;
+  }
+
+  try {
+    // Use as any to bypass type checking for the config if necessary,
+    // but the config should be valid after trimming keys.
+    const messages = linter.verify(code, config as any, {
+      filename: 'file.js',
```
curran, line 79: We could potentially pass the correct file name in here. I'm not sure where that is surfaced later, but might be a good idea in case this file name is used in the UI or messages.

src/client/useESLint/worker.ts
```diff
@@ -0,0 +1,120 @@
+import * as eslint from 'eslint-linter-browserify';
+import globals from 'globals';
+
+const linter = new eslint.Linter();
+
+// Helper function to trim keys of an object
+function trimGlobalKeys(
```
curran, line 7: Why is this needed?

I don't see any keys here that have leading or trailing white space:

https://github.com/sindresorhus/globals/blob/main/globals.json

I would suggest to remove this logic if there's no clear reason to have it.

src/client/useESLint/worker.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+  return newObj;
+}
+
+// Define the configuration without explicit types that don't exist
+const config = {
+  languageOptions: {
+    globals: {
+      ...trimGlobalKeys(globals.browser), // Apply trimming to browser globals
```
curran, line 26:
```suggestion
      ...globals.browser
```

src/client/useESLint/index.ts
```diff
@@ -0,0 +1,64 @@
...
+// @ts-ignore - Worker import
+import ESLintWorker from './worker?worker';
+import { enableESLint } from '../../server/featureFlags';
+import { fileNameStateField } from '../CodeEditor/getOrCreateEditor';
+
+let requestCounter = 0;
+const pendingRequests = new Map();
+
+export const useESLint = () => {
+  const [worker] = useState(() => new ESLintWorker());
```
curran, line 13: Nice pattern! I have never seen this before.

src/client/useESLint/worker.ts
```diff
@@ -0,0 +1,120 @@
...
+  },
+  rules: {
+    'no-unused-vars': ['warn', { argsIgnorePattern: '^_' }],
+    'no-undef': 'error',
+    semi: 'off',
+  },
+};
+
+// Helper function to convert line/column to character offset
+function getOffset(
```
curran, line 46: I'm a bit confused on why this logic is needed, because in the original example here

https://github.com/UziTech/eslint-linter-browserify/blob/master/example/script.js

there is no additional logic for conversions like this. I'm curious, is this logic coming from some other reference example you are looking at as an example?

curran (changes requested):
This is amazing! Thank you so much for this contribution.

I left some comments inline.

curran:
Maybe have a quick look @Darmin-Geek . Thanks!

src/client/useESLint/worker.ts
fantong11: ESLint only gives you "line + column" instead of a "string position", so I used a offset function to handle

https://codemirror.net/examples/lint/

package.json
```diff
@@ -156,6 +159,7 @@
     "sharedb-client-browser": "^5.1.1",
     "uuid": "^11.1.0",
     "vizhub-ui": "^3.23.0",
+    "vzcode": "file:",
```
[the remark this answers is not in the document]
fantong11: My vscode extension settings have some problem, it automatically add the dependencies.

package.json
```diff
@@ -172,7 +176,6 @@
     "vitest": "^3.1.1"
   },
   "optionalDependencies": {
-    "@rollup/rollup-darwin-arm64": "^4.39.0",
-    "@rollup/rollup-win32-x64-msvc": "^4.39.0"
```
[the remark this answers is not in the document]
fantong11: vscode extension settings issue, too.

package.json
```diff
@@ -172,7 +176,6 @@
...
+    "@rollup/rollup-darwin-arm64": "^4.39.0"
```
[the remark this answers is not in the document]
fantong11: vscode extension settings issue.

src/client/useESLint/worker.ts
fantong11: Added

src/client/useESLint/worker.ts
fantong11: I will remove this logic.
This piece of code was accidentally added because I was using AI. 
I thought this code is necessary.

fantong11:
I’ve made some minor fixes to the code you thought was problematic.

src/client/useESLint/worker.ts
fantong11: This option is not necessary, I will remove.

curran (approved):
LGTM! Thank you so much.

PR #918, opened 2025-05-16
844 adopt val towncodemirror ts for typescript support

### LINKED ISSUE #844: Adopt val-town/codemirror-ts for TypeScript support

Darmin-Geek:
This PR adds type hints and error information on hover. It also replaces the existing autocomplete such that matching types are recommended first. 

NPM_Based_valtown_1

Unfortunately, ATA, which adds types for NPM packages, is not supported in the current release of codemirror-ts on NPM. As a result, I did not include it in this PR. Let me know if I should include ATA, noting that it currently only supports NPM packages. 

Also, typescript detects an import error that does not surface when the typescriptExtension/worker.ts file is used as a web worker. worker.ts is only used as a web worker. To avoid future confusion, I have suppressed the error.

Closes #844


curran:
Thank you so much! I will review this as soon as I get a chance.

src/client/CodeEditor/getOrCreateEditor.ts
```diff
@@ -41,17 +41,38 @@ import { ThemeLabel, themeOptionsByLabel } from '../themes';
...
+  tsSyncWorker,
+} from '@valtown/codemirror-ts';
 
 const DEBUG = false;
 
 // Enables TypeScript +JSX support in CodeMirror.
 const tsx = () =>
   javascript({ jsx: true, typescript: true });
 
+const innerWorker = new Worker(
```
curran, line 64: This will likely fail in an SSR context, which is there in the VizHub app but not in the VZCode demo app (so it works fine here). We can circle back and fix this later.

PR #920, opened 2025-05-18
Clicking on search result leads to VZCode opening the file and scrolling to the relevant text

Darmin-Geek:
Pressing enter on a search result leads to the same outcome.


curran (approved):
LGTM! Thank you.

PR #925, opened 2025-06-06
#921 es lint for jsx files ethan tamang

### LINKED ISSUE #921: ESLint for JSX Files
Image

As a follow-up to

 * https://github.com/vizhub-core/vzcode/issues/915

There seems to be an issue where perhaps the JSX parsing may not be happening as expected. In the screenshot above, it says `App` is not used, but in fact it is used, just in JSX.

```jsx
import { createRoot } from 'react-dom';
// This variable is used, but ESLint thinks it is not used
import App from './app';

const root = createRoot(document.getElementById('root'));
root.render(<App />);
```

TmgEE:
Closes #921

some progress with plugin


curran:
Some feedback:

 * Great direction! Let's keep going with the server-side approach
 * The alternative would be to use Browserify to create a browser build of https://www.npmjs.com/package/eslint-plugin-react

curran:
@TmgEE How is this going? Is it ready for merge or more work needed? Thanks!

PR #939, opened 2025-07-19
created-new-AI-chat-bot-icon

nitanagdeote:
image


curran:
Great idea to improve the icon!

This is not quite ready, but I like the direction.

Maybe we could make it look like the "VizBot" PoC image you made.

curran:
It's good idea! Let's make an icon for "VizBot".

image

PR #941, opened 2025-07-19
Move ai chat to landing page

nitanagdeote:


curran:
Good direction, but the approach is not quite right. This repo contains just the editor, with no concept of pages. The landing page lives in the vizhub main repo, over in https://github.com/vizhub-core/vizhub. That's where all the various "Pages" are defined.

There's already a PoC page that exists for doing this, which I think your idea is to put the box like Bolt.new where you just enter a prompt, and it automatically forks a pre-selected viz and applies the AI edit. Right?

```
vizhub/api/src/endpoints/createVizFromPromptEndpoint.ts
```

The PoC for that is here:

```
vizhub/app/src/pages/AILandingPage/PromptPage.tsx
```

We're tracking this in the issue:

 * https://github.com/vizhub-core/vizhub/issues/1098

PR #942, opened 2025-07-19
made the AI-chat-Better

nitanagdeote:
image


curran:
Nice! I like the little green dot, and I like the colors and gradient change.

Unfortunately this PR breaks everything, so it's not ready for merge, but some of the ideas are good. Maybe try for a new PR that just makes a small change, e.g. adding the gradient, or adding the text styles, or adding the green circle?

This is what I see when I test this out locally:

image

 * The AI chat is not in the right place (it's to the left of the icon list)
 * There is another AI chat area to the right, which looks broken
 * The AI chat is always visible, but should not be

PR #1185, opened 2025-11-22
Reveal model

curran:
It shows the model now:

image


curran:
Status:

 * Problems around duplicated code
 * Resolution: delete the unused ones

